PR #271, opened 2017-08-15
PP-2502 Apply custom branding in Frontend

kakumara:
## WHAT
This PR, hooks a `Service's` custom barding in to our payment pages. 
The wiring with backend in done using;
 - using the `views.js` to declare `customBrandingCssPath` and `customBrandingImagePath` template variables which is used to override GOV.UK Pay branding from our payment pages.
- In order to resolve the `Service` for the `Charge` in context, I've introduced a middleware `resolve_service.js` and applied on the payment routes. 

Regarding styling bit ( mimicing @jonheslop )
 - Had to override GOV.UK templates with a monkey patched version (see `package.json` `govuk_template_mustache_inheritance`) in order to override fonts options properly
(after discussing with GOV.UK template folks)

Other than that had to further extend `charge_ft_test.js` in order to make it work with the new middleware. This looks 😖 but leaving the refactoring for another day.


app/utils/views.js
```diff
@@ -205,16 +205,22 @@ module.exports = (function () {
...
+        locals.customBrandingCssPath = _.get(res.locals, 'service.customBranding.cssUrl')
+        locals.customBrandingImagePath = _.get(res.locals, 'service.customBranding.imageUrl')
+      }
+
       if (_.get(locals, 'analytics.path')) {
         locals.analytics.path = locals.analytics.path + _.get(action, 'analyticsPage', '')
       }
       status = (action.code) ? action.code : 200
       res.status(status)
+      console.log(`${action.view} ... ${JSON.stringify(locals)}`)
```
mrlumbu: remove this I guess

kakumara: oops 🙈

package.json
```diff
@@ -50,7 +50,7 @@
     "express": "4.15.x",
     "forever": "0.15.x",
     "govuk_frontend_toolkit": "6.0.x",
-    "govuk_template_mustache": "0.22.x",
+    "govuk_template_mustache_inheritance": "https://github.com/jonheslop/govuk_template/releases/download/v0.22.4/mustache_inheritance_govuk_template-0.22.3.tgz",
```
mrlumbu: so is this just a thing that jon's stuck on his personal github? Do we have a plan for doing something a bit more sustainable?

mrlumbu:
Think I'm happy for this to be merged - I have put a few questions in but none are blockers (but perhaps should be considered in subsequent PRs). 
One other thing I wanted to check is the general approach, and who has been involved in kicking this off so far: the approach as it stands (and as I understand it) gives a service a lot of power in that they can specify a css url, and we will happily load that in? What thought are we giving to CSS injection etc? Or are we just trusting services to 'be good'?

app/middleware/resolve_service.js
```diff
@@ -0,0 +1,23 @@
...
+
+module.exports = function (req, res, next) {
+  'use strict'
+  const _views = views.create()
+
+  if (!req.chargeId && !req.chargeData) {
+    _views.display(res, 'UNAUTHORISED', withAnalyticsError())
+  } else {
+    let correlationId = req.headers[CORRELATION_HEADER]
+    return getAdminUsersClient({correlationId: correlationId}).findServiceBy({gatewayAccountId: req.chargeData.gateway_account.gateway_account_id})
```
kakumara: If that's not present that'll be a bigger problem isn't it? Having a charge without a gateway account id? Sounds like a impossible scenario, but I'll surround it using `lodash`.get so at least there won't be referencing errors if that ever happens.

app/utils/base_client2.js
```diff
@@ -19,17 +19,14 @@ function retryOnEconnreset (err) {
...
-
 if (process.env.DISABLE_INTERNAL_HTTPS !== 'true') {
-  httpsAgent.ca = customCertificate.getCertOptions()
+  agentOptions.ca = customCertificate.getCertOptions()
 } else {
   logger.warn('DISABLE_INTERNAL_HTTPS is set.')
 }
 
+const httpsAgent = new https.Agent(agentOptions)
+
```
kakumara: I tried to get rid of baseClient but seems like a bigger change. Hence the `base_client2.js` which is the same supporting `pact` stuff. This is new and was not used before. On my previous commit I had this options setting up wrong way around.

package.json
kakumara: to be honest I didn't fully understand this fully either. But @jonheslop seemed to have had a chat with GOV.UK folks about this and they have confirmed that this is the right way for now.

Although, I still cannot justify why its in his personal repo (just realised when you mentioned it). He could have easily done this some where in `alphagov`. I'll chase it with him when he's back

PR #275, opened 2017-08-18
PP-2502 including the non-GOV.UK branding changes with the layout fixes

kakumara:
These were reverted temporarily to fix a issue on display titles


app/views/layout.html
```diff
@@ -5,10 +5,25 @@
...
+{{$fontsCSS}}
+{{/fontsCSS}}
+{{/customBrandingImagePath}}
+{{/insideHeader}}
+
 {{$propositionHeader}}
 {{>includes/propositional_navigation}}
 {{/propositionHeader}}
 
+{{$bodyClasses}}custom-branding{{/bodyClasses}}
```
jonheslop: @kakumara This is the line that is causing issues I think. this needs to be conditional

PR #277, opened 2017-08-31
PP-2502 - Finishing up the logic to enable custom branding

jonheslop:
Moving things around to be neater and added in logic to enable custom branding when it's been set.

We had to reset the fonts it loads which sorta sucks but it's the only way around it as mustache doesn't let you write logic the other way around 🙄

### DIFF
--- a/app/utils/views.js
+++ b/app/utils/views.js
@@ -211,8 +211,7 @@ module.exports = (function () {
       locals = (action.locals) ? _.merge({}, action.locals, locals) : locals
 
       if (_.get(res.locals, 'service.hasCustomBranding', false)) {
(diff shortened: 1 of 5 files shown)


app/utils/views.js
```diff
@@ -211,8 +211,7 @@ module.exports = (function () {
       locals = (action.locals) ? _.merge({}, action.locals, locals) : locals
 
       if (_.get(res.locals, 'service.hasCustomBranding', false)) {
-        locals.customBrandingCssPath = _.get(res.locals, 'service.customBranding.cssUrl')
-        locals.customBrandingImagePath = _.get(res.locals, 'service.customBranding.imageUrl')
+        locals.customBranding = _.get(res.locals, 'service.customBranding')
```
kakumara: 👍

PR #285, opened 2017-09-11
[DO NOT MERGE] PP-2617 Upgraded charge controller to use new base client

brid:
## WHAT
Upgrades all outbound http operation to upstream/downstream services to utilise the new base_client which derives from the npm request module, rather than our original pattern of calling the raw http/https modules directly.


kakumara:
Thanks for doing this. Didn't have time to get it working at the time

PR #301, opened 2017-10-13
Remove unused dependencies

mrlumbu:
Remove array.prototype.find and string from dependencies, 
as they are not used.


mrlumbu:
😀

PR #308, opened 2017-10-27
PP-2688 Add merchant details

rauligs:
## WHAT
- Add merchant details above privacy info in footer area.
- Extract privacy info into a separated template.
- Add merchant details to Service object class **when available.** Added _isEmpty_ method to check whether merchant details exist so it is backward compatible (current services do not include merchant details, a test account might not have them but live account must have merchant details available).

### DIFF
--- a/app/models/Service.class.js
+++ b/app/models/Service.class.js
@@ -16,7 +16,6 @@ class Service {
    * @returns {Object} service - internal representation of the service object
    **/
   constructor (serviceData) {
-    console.log(serviceData.merchant_details.address_country)
     this.externalId = serviceData.external_id
     this.name = serviceData.name
     this.gatewayAccountIds = serviceData.gateway_account_ids
--- a/app/models/Service.class.js
+++ b/app/models/Service.class.js
@@ -39,7 +39,20 @@ class Service {
    * @returns {boolean} if the service got a non-GOV.UK branding
    */
   hasCustomBranding () {
-    return !_.isEmtpy(this.customBranding)
+    return !_.isEmpty(this.customBranding)
+  }
+
+  /**
+   * @method hasMerchantDetails
+   *
+   * All services must have merchant details defined for all gateway accounts
+   * This will be kept as backward compatibility until merchant details are
+   * enforced.
+   *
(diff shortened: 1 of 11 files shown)


app/models/Service.class.js
```diff
@@ -51,6 +41,19 @@ class Service {
...
+   * @method hasMerchantDetails
+   *
+   * All services must have merchant details defined for all gateway accounts
+   * This will be kept as backward compatibility until merchant details are
+   * enforced.
+   *
+   * @returns {boolean} if the service got merchant details specified
+   */
+  hasMerchantDetails () {
+    return !_.isEmtpy(this.merchantDetails)
```
mrlumbu: isEmpty

test/models/service_test.js
```diff
@@ -0,0 +1,39 @@
...
+        address_line2: '10 Downing Street',
+        address_city: 'London',
+        address_postcode: 'AW1H 9UX',
+        address_country: 'GB'
+      }
+    }
+
+    let serviceModel = new Service(data)
+
+    assert.equal(serviceModel.merchantDetails.countryName, 'United Kingdom')
```
mrlumbu: think we're trying to standardise on `expect` no? Check with @maxcbc

test/models/service_test.js
```diff
@@ -0,0 +1,39 @@
...
+    let data = {
+      external_id: '1234',
+      name: 'service name',
+      gateway_account_ids: [1],
+      custom_branding: {css_url: 'css url', image_url: 'image url'}
+    }
+
+    let serviceModel = new Service(data)
+
+    assert.equal(serviceModel.merchantDetails, undefined)
```
mrlumbu: likewise

test/models/service_test.js
maxcbc: Yeah. We've been trying to standardise on chai's expect style. Currently we're using at least 3 different styles across the projects.

Happy to help out if you're unfamiliar with `expect`.

maxcbc:
Love that you're testing things to the level of class methods. Many of these -> 👍 👍 👍 👍 👍

PR #312, opened 2017-11-06
PP-2688 Merchant details CSS changes

DanailMinchev:
## WHAT

Use `display: block` to display the merchant details name and address on two lines when `address line 2` is not provided.

with @jonheslop

### DIFF
--- a/app/assets/sass/pages/card-details.scss
+++ b/app/assets/sass/pages/card-details.scss
@@ -59,3 +59,11 @@
   cursor: pointer;
   font-size: 19px;
 }
+
+#footer .footer-meta .footer-meta-inner {
+  .merchant-details,
+  .merchant-details-line-1,
+  .merchant-details-line-2 {
+    display: block;
+  }
+}
--- a/app/views/includes/merchant_details.html
+++ b/app/views/includes/merchant_details.html
@@ -1,5 +1,5 @@
 {{#merchantDetails}}
-<ul>
+<ul class="merchant-details">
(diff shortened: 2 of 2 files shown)


app/assets/sass/pages/card-details.scss
```diff
@@ -59,3 +59,11 @@
   cursor: pointer;
   font-size: 19px;
 }
+
+#footer .footer-meta .footer-meta-inner {
+  .merchant-details,
+  .merchant-details-line-1,
```
simad: have you tried to assign the `display: block` only to merchant-details? It should be inherited by the `<li>`s, so you might be able to simplify the code

jonheslop: @simad it's needs to be this specific to override the really specific code coming from govuk_template 🙄

DanailMinchev: Thanks @simad .
I've tried that initially, but unfortunately it doesn't work, because we have more specific rule `#footer .footer-meta .footer-meta-inner ul li` in [_footer.scss#L68](https://github.com/alphagov/govuk_template/blob/e9ae7ff045e0adf552aeed509be5ef174be325aa/source/assets/stylesheets/_footer.scss#L68) which overrides less specific one: `#footer .footer-meta .footer-meta-inner .merchant-details`

simad: gotcha, thanks

DanailMinchev:
Closing this one, because Jenkins and GitHub integration is having problems

PR #317, opened 2017-11-17
Use typographic apostrophe in email field hint

quis:
> “Smart quotes,” the correct quotation marks and apostrophes, are curly or sloped. "Dumb quotes," or straight quotes, are a vestigial constraint from typewriters when using one key for two different marks helped save space on a keyboard. Unfortunately, many improper marks make their way onto websites because of dumb defaults in applications and CMSs.

– http://smartquotesforsmartpeople.com/

This commit replaces an instance of a straight single quote used as an apostrophe (') with the unicode character for a proper typographic apostrophe (’).

I haven’t looked for other occurrences of this happening on this page or in this repo.

### DIFF
--- a/app/views/charge.html
+++ b/app/views/charge.html
@@ -271,7 +271,7 @@ Enter your card details.
                 data-original-label="{{ i18n.chargeView.email }}">
                 {{ i18n.chargeView.email }}
               </span>
-              <span class="form-hint">We'll send your payment confirmation here</span>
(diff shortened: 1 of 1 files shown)


jonheslop:
@quis I think our Jenkins bot won’t build ‘n‘ test this PR as it doesn’t recognise you as a valid contributor 😞 (PCI and all that jazz)

I’ll close this PR and replicate myself.

PR #318, opened 2017-11-20
PP-2915 - increased border width inline with GOV.UK style

jonheslop:
Borders were thinner than GOV.UK standard style 

Also fixed a bunch of styelint errors

### DIFF
--- a/app/assets/sass/forms/_form-errors.scss
+++ b/app/assets/sass/forms/_form-errors.scss
@@ -16,7 +16,7 @@
 
   // Form inputs should have a red border
   .form-control {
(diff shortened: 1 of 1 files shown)


app/assets/sass/forms/_form-errors.scss
```diff
@@ -16,7 +16,7 @@
 
   // Form inputs should have a red border
   .form-control {
-    border-color:$error-colour;
+    border: 4px solid $error-colour;
```
jonheslop: Meaningful change, everything else is just the linter reordering

PR #319, opened 2017-11-21
PP-1797 - Added an .editorconfig to keep things consistent

jonheslop:
EditorConfig helps developers define and maintain consistent coding styles between different editors and IDEs. The EditorConfig project consists of a file format for defining coding styles and a collection of text editor plugins that enable editors to read the file format and adhere to defined styles.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*.{js,json,scss}]
+charset = utf-8
+end_of_line = lf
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
+
+[*]
```
DanailMinchev: Should we limit this to `[*.{js,json}]` ?

PR #326, opened 2017-12-02
PP-2992 Nunjucks migration

DanailMinchev:
## WHAT

We’ve decided to migrate from Mustache to [Nunjucks](https://mozilla.github.io/nunjucks/) because Mustache is no longer maintained or used by the Node.js community, Nunjucks on the other hand is widely used. It’s also used widely within GDS (and also shares it's syntax with Jinja2 which GDS Python projects use)

## HOW 

This PR includes the migration to Nunjucks as well as tests improvements [c5ac7f3] : before we were testing against JSON template data and now we are testing against generated HTML


server.js
```diff
@@ -1,103 +1,189 @@
...
+    autoescape: true, // controls if output with dangerous characters are escaped automatically
+    throwOnUndefined: false, // throw errors when outputting a null/undefined value
+    trimBlocks: true, // automatically remove trailing newlines from a block/tag
+    lstripBlocks: true, // automatically remove leading whitespace from a block/tag
+    watch: false, // reload templates when they are changed (server-side). To use watch, make sure optional dependency chokidar is installed
+    noCache: false // never use a cache and recompile templates each time (server-side)
   }
-  res.locals.session = function () {
-    return session.retrieve(req, req.chargeId)
+  if ((!NODE_ENV) || (NODE_ENV !== 'production')) {
```
maxcbc: This could be inlined;
```
watch: NODE_ENV !== 'production',
noCache: NODE_ENV !== 'production'
```

That means you could inline both `APP_VIEWS` and `nunjucksConfiguration`

DanailMinchev: Thanks, Max.

I declared the `if ((!NODE_ENV) || (NODE_ENV !== 'production'))` block in case we don't have `NODE_ENV` defined (in dev using `msl`) and also in case we may need other `non-boolean` properties for non-prod declarations as in https://github.com/alphagov/pay-products-ui/blob/d157974c7ddacb834b9f43e6e0e31653e9544652/server.js#L83.
I can change it like:
```
watch: ((!NODE_ENV) || (NODE_ENV !== 'production')),
noCache: ((!NODE_ENV) || (NODE_ENV !== 'production'))
```
adding extra parentheses for more clarity.
Let me know what you prefer.

package.json
```diff
@@ -47,9 +47,8 @@
     "csrf": "3.0.x",
     "express": "4.15.x",
     "forever": "0.15.x",
-    "govuk_frontend_toolkit": "6.0.x",
-    "govuk_template_mustache_inheritance": "https://github.com/jonheslop/govuk_template/releases/download/v0.22.4/mustache_inheritance_govuk_template-0.22.3.tgz",
-    "hogan.js": "3.0.x",
+    "govuk_frontend_toolkit": "7.2.0",
+    "govuk_template_jinja": "^0.23.0",
```
jonheslop: Need to use:
`    "govuk_template_jinja": "https://github.com/jonheslop/govuk_template/releases/download/v0.22.4/jinja_govuk_template-0.22.3.tgz",`

DanailMinchev: Thanks, Jon

Addressed the change

test/integration/charge_validation_ft_tests.js
```diff
@@ -55,15 +54,25 @@ describe('checks for PAN-like numbers', () => {
...
-        expect(body.highlightErrorFields.email).to.equal('Enter a valid email')
+        const $ = cheerio.load(res.text)
+        const errorMessages = {
+          cardholder: 'Enter the name as it appears on the card',
+          addressLine1: 'Enter a valid billing address',
+          city: 'Enter a valid town/city',
+          postcode: 'Enter a valid postcode',
+          email: 'Enter a valid email'
+        }
+        expect($('#cardholder-name-error').text()).to.contains(errorMessages.cardholder)
```
maxcbc: This looks great Dan, fantastic work

package.json
```diff
@@ -47,11 +47,13 @@
     "csrf": "3.0.x",
     "express": "4.16.x",
     "forever": "0.15.x",
-    "hogan.js": "3.0.x",
+    "govuk_frontend_toolkit": "7.2.0",
```
maxcbc: should be devDeps

package.json
```diff
@@ -47,11 +47,13 @@
...
+    "govuk_template_jinja": "https://github.com/jonheslop/govuk_template/releases/download/v0.22.4/jinja_govuk_template-0.22.3.tgz",
```
maxcbc: should be devDeps

PR #348, opened 2018-01-29
PP-3241 Incorrect page titles on frontend

belindac:
Minor text changes requested to the charge and confirm pages so that the
title of the page matches the page heading.

Goes with https://github.com/alphagov/pay-accepttests/pull/195

@solo


belindac:
After speaking about this change, it was decided that we should not be asserting on titles, so the check in acceptance tests will be removed

PR #351, opened 2018-02-01
BAU Minor Refactoring

maxcbc:
# BAU Minor Refactoring

## What
* 7e62443 - Usage of `module.exports.<property-name>` removed in favour of `exports.<property-name>`. 
*  56235c0 - Usage of `require(path.join(__dirname, '<file-path>'))` removed in favour of `require('<file-path>')`

## Why
* The variable `exports`, when a module is instantiated is a shorthand reference to the `exports` property of the module. i.e. `exports = module.exports = {}`
* `require()` in Node.js handles relative file loading by default so using `path.join(__dirname, '<file-path>') is unnecessary.

## More detailed why
* It's generally considered better practice to:
    * use `exports.<property-name>` when assigning a property to be exported
    * use `module.exports = <Object | Function>` when exporting a single export.
* Using `require(path.join(__dirname, '<file-path>'))` in JetBrains IDEs breaks automatic updates to file paths when moving files, meaning a developer has to manually update each instance of a file being imported into another module. This is exceptionally annoying, especially if you are attempting any sort of refactoring.

### DIFF
--- a/app/browsered.js
+++ b/app/browsered.js
@@ -1,3 +1,3 @@
 window.jQuery = window.$ = require('jquery')
-module.exports.chargeValidation = require('./utils/charge_validation')
-// module.exports.accessibleAutocomplete = require('./utils/accessible-autocomplete.js')
+exports.chargeValidation = require('./utils/charge_validation')
+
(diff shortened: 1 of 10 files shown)


maxcbc:
Closed as has now been assigned Jira issue

PR #355, opened 2018-02-02
PP-3266 refactor return secure and static controllers

maxcbc:
## WHAT
In the secure, return and static controllers:
- removed unnecessary inner functions in secure controller
- removed unnecessary variables in return controller
- implemented 'prefer-const' style for variable declarations

In the secure controller:
- fixed line 20 (old line 28) where the success path would be executed even if the token destroy failed.
- fixed line 20 (old line 28) where the error would be lost to the ether if token destroy failed.

### DIFF
--- a/app/controllers/return_controller.js
+++ b/app/controllers/return_controller.js
@@ -1,12 +1,17 @@
-var Charge = require('../models/charge.js')
-var StateModel = require('../models/state.js')
-var views = require('../utils/views.js')
-var CORRELATION_HEADER = require('../utils/correlation_header.js').CORRELATION_HEADER
-var withAnalyticsError = require('../utils/analytics.js').withAnalyticsError
+'use strict'
 
-var logger = require('winston')
+// npm dependencies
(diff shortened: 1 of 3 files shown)


app/controllers/return_controller.js
```diff
@@ -1,12 +1,17 @@
...
-var CORRELATION_HEADER = require('../utils/correlation_header.js').CORRELATION_HEADER
-var withAnalyticsError = require('../utils/analytics.js').withAnalyticsError
+'use strict'
 
-var logger = require('winston')
+// npm dependencies
+const logger = require('winston')
 
-let CANCELABLE_STATES = [
+// local dependencies
```
jonheslop: This is fine for now, but if we move to XO it will want `// Local dependencies`

jonheslop: But TBF if we do change `xo --fix` would sort it out

maxcbc: lol

PR #359, opened 2018-02-07
PP-3266 refactor middleware

maxcbc:
## WHAT
Refactor middleware to:
 - use strict
 - prefer const
 - eliminate unnecessary inner functions and variables
 - to use `exports.<property-name>` over `module.exports.<property-name>`

### DIFF
--- a/app/middleware/csrf.js
+++ b/app/middleware/csrf.js
@@ -1,77 +1,44 @@
-var csrf = require('csrf')
-var session = require('../utils/session.js')
-var views = require('../utils/views.js')
-var chargeParam = require('../services/charge_param_retriever.js')
-var logger = require('winston')
-var _views = views.create()
+'use strict'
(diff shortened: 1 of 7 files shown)


app/middleware/csrf.js
```diff
@@ -1,77 +1,44 @@
...
-var chargeParam = require('../services/charge_param_retriever.js')
-var logger = require('winston')
-var _views = views.create()
-
-var csrfTokenGeneration = function (req, res, next) {
-  'use strict'
-  var chargeId = chargeParam.retrieve(req)
-  var chargeSession = session.retrieve(req, chargeId)
-
-  var init = function () {
```
jonheslop: I see what you were talking about now these inits are silly!

maxcbc: true dat

PR #361, opened 2018-02-08
PP-3266 refactor validators

maxcbc:
## WHAT
- Refactor validators to:
    - use strict
    - prefer const
    - eliminate unnecessary inner functions and variables
    - to use exports.<property-name> over module.exports.<property-name>
 - Update babel implementation to:
    - target our target browser range
    - use polyfills


Gruntfile.js
```diff
@@ -135,7 +135,13 @@ module.exports = function (grunt) {
         [
           'babelify',
           {
-            presets: ['es2015']
+            presets: [
+              ['env', {
+                'targets': {
+                  'browsers': ['last 2 versions', 'safari >= 7', 'ie >= 10']
```
jonheslop: 👌

app/browsered.js
```diff
@@ -1,2 +1,3 @@
+require('babel-polyfill')
```
jonheslop: 🤦‍♂️

maxcbc: lol

PR #365, opened 2018-02-15
PP-3302 accessible autocomplete

jonheslop:
## 3d12fcb PP-3302 - Installing autocomplete module 
I've used [govuk-country-and-territory-autocomplete](https://github.com/alphagov/govuk-country-and-territory-autocomplete) and not accessible-autocomplete. As in this case we are using it to pick countries and also it has a bunch of neat features layered on top which make it a better autocomplete.

## e06905f PP-3302 - Integrate autocomplete 
- Using countries json from the module and removing old list.
- Rewritten services/countries.js to use the new list
- Renamed function translateAlpha2 to clear name translateCountryISOtoName
- Pulled assets from module and bundled them in with Grunt
- Noticed we weren't using minified CSS, so switched to that
- Updated tests to reflect changes


app/services/countries.js
```diff
@@ -4,22 +4,7 @@
 const lodash = require('lodash')
 
 // Local Dependencies
-let countries = require('../data/countries.json')
-const extensions = require('../data/country-record-extension.json')
+let countries = require('../../node_modules/govuk-country-and-territory-autocomplete/dist/location-autocomplete-canonical-list.json')
```
maxcbc: should be `const`

app/services/countries.js
```diff
@@ -4,22 +4,7 @@
...
 
-// Exports
 exports.retrieveCountries = () => lodash.clone(countries)
```
maxcbc: why do we need this?

PR #377, opened 2018-03-19
Pp 3453 handle epdq 3ds response

georgeracu:
## WHAT
- When receiving a response from the payment gateway, it will
send the user back inside the iframe. To get out from the iframe
we use a form with a target `_top` and at the same time we pass
any data we get from the payment provider as form fields to the
charge_controller. Because some of this data is specific to each
payment gateway, we will set it only if available. After setting this
data as hidden fields we then submit the form and send this data
to connector
- Added both POST and GET to the same path for EPDQ as they
use GET as a fall back when JS is deactivated. This callback url to us
will contain the EPDQ auth status as query param
- The status received from EPDQ is not reliable as this is
resubmitted as a hidden field, however we verify this status in connector
against EPDQ before updating the charge

## HOW 
- Updated auth_3ds template to hide/display data for EPDQ or
WorldPay
- Updated routes for handling EPDQ 3DS

### DIFF
--- a/app/controllers/charge_controller.js
+++ b/app/controllers/charge_controller.js
@@ -55,6 +55,21 @@ function redirect (res) {
   }
 }
 
+function build3dsPayload (req) {
+  let auth3dsPayload = {}
+  const paRes = _.get(req, 'body.PaRes')
+  if (!_.isUndefined(paRes)) {
+    auth3dsPayload.pa_response = paRes
+  }
+
+  const providerStatus = AUTH_3DS_EPDQ_RESULTS[_.get(req, 'body.providerStatus', '')]
+  if (!_.isUndefined(providerStatus)) {
+    auth3dsPayload.auth_3ds_result = providerStatus
+  }
+
+  return auth3dsPayload
+}
+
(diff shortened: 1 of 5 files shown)


app/controllers/charge_controller.js
```diff
@@ -147,8 +152,11 @@ module.exports = {
     const startTime = new Date()
     const correlationId = req.headers[CORRELATION_HEADER] || ''
     const connector3dsUrl = paths.generateRoute('connectorCharge.threeDs', {chargeId: charge.id})
-
-    baseClient.post(connector3dsUrl, {data: {pa_response: _.get(req, 'body.PaRes')}, correlationId},
+    const auth3dsPayload = {
+      pa_response: _.get(req, 'body.PaRes'),
+      auth_3ds_result: AUTH_3DS_EPDQ_RESULTS[_.get(req, 'body.providerStatus', '')]
```
ckalista: If it is none of the `AUTH_3DS_EPDQ_RESULTS` the app would crash. Is that expected behaviour?

PR #382, opened 2018-03-21
PP-3438 Remove products smoke test from front end build

mrlumbu:
Card smoke test tests all relevant functionality from a front
end point of view, so we don't need an additional smoke test.


tlwr:
Closing as already done

PR #386, opened 2018-03-26
PP-3528 Enable 3DS for SmartPay

georgeracu:
## WHAT
- Enable 3DS for SmartPay

## HOW 
- Added MD field to the form
- This extra field is required by SmartPay 3DS

### DIFF
--- a/app/controllers/charge_controller.js
+++ b/app/controllers/charge_controller.js
@@ -206,19 +206,16 @@ module.exports = {
     const md = _.get(charge, 'auth3dsData.md')
     const htmlOut = _.get(charge, 'auth3dsData.htmlOut')
 
-    if (issuerUrl && paRequest && md) {
-      views.display(res, AUTH_3DS_REQUIRED_OUT_VIEW, {
-        issuerUrl: issuerUrl,
-        paRequest: paRequest,
-        md: md,
-        threeDSReturnUrl: `${req.protocol}://${req.hostname}${paths.generateRoute('external.card.auth3dsRequiredIn', {chargeId: charge.id})}`
-      })
-    } else if (issuerUrl && paRequest) {
-      views.display(res, AUTH_3DS_REQUIRED_OUT_VIEW, {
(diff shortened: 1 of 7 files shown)


app/controllers/charge_controller.js
```diff
@@ -210,6 +216,13 @@ module.exports = {
       views.display(res, AUTH_3DS_REQUIRED_HTML_OUT_VIEW, {
         htmlOut: Buffer.from(htmlOut, 'base64').toString('utf8')
       })
+    } else if (issuerUrl && paRequest && md) {
```
SandorArpa: You'll never arrive here!

app/controllers/charge_controller.js
```diff
@@ -198,9 +203,17 @@ module.exports = {
     const charge = normalise.charge(req.chargeData, req.chargeId)
     const issuerUrl = _.get(charge, 'auth3dsData.issuerUrl')
     const paRequest = _.get(charge, 'auth3dsData.paRequest')
+    const md = _.get(charge, 'auth3dsData.md')
     const htmlOut = _.get(charge, 'auth3dsData.htmlOut')
 
-    if (issuerUrl && paRequest) {
+    if (issuerUrl && paRequest && md) {
```
kakumara: This could be simplified to be something like this;
```
if (issuerUrl && paRequest) {
      let data = {
        issuerUrl: issuerUrl,
        paRequest: paRequest,
        threeDSReturnUrl: `${req.protocol}://${req.hostname}${paths.generateRoute('external.card.auth3dsRequiredIn', {chargeId: charge.id})}`
      }
      if (md) {
        data.md = md
      }
      views.display(res, AUTH_3DS_REQUIRED_OUT_VIEW, data)
    }
```

kakumara:
I can only see a test for the `in` flow. Shouldn't there be a test for the out flow as well? (to see if `md` is rendered in the `out` html ?

PR #390, opened 2018-04-03
PP-3579 - AB testing - Firebreak

tlwr:
# A/B Testing

- Added middleware to split traffic between two routes
- Added analytics so we can measure which one they used
- Amended tests and set up env vars and query params so we can disable if we want to

### DIFF
--- a/app/router.js
+++ b/app/router.js
@@ -14,6 +14,8 @@ const stateEnforcer = require('./middleware/state_enforcer.js')
 const retrieveCharge = require('./middleware/retrieve_charge.js')
 const resolveService = require('./middleware/resolve_service.js')
 
+const abTest = require('./utils/ab_test.js')
+
 exports.paths = paths
 
 exports.bind = function (app) {
@@ -41,7 +43,15 @@ exports.bind = function (app) {
     stateEnforcer
   ]
 
-  app.get(card.new.path, middlewareStack, charge.new)
+  app.get(
+    card.new.path,
+    middlewareStack,
+    abTest.switch({
+      threshold:      90,
+      defaultVariant: charge.new,
+      testingVariant: charge.new
+    })
+  )
   app.get(card.authWaiting.path, middlewareStack, charge.authWaiting)
   app.get(card.auth3dsRequired.path, middlewareStack, charge.auth3dsRequired)
   app.get(card.auth3dsRequiredOut.path, middlewareStack, charge.auth3dsRequiredOut)
--- a/app/router.js
+++ b/app/router.js
@@ -58,7 +58,7 @@ exports.bind = function (app) {
     card.confirm.path,
     middlewareStack,
     abTest.switch({
-      threshold: 90,
+      threshold: process.env.AB_TEST_THRESHOLD || 100,
       defaultVariant: charge.confirm,
       testingVariant: charge.confirmVariant
     })
--- a/app/router.js
+++ b/app/router.js
@@ -13,8 +13,9 @@ const actionName = require('./middleware/action_name.js')
 const stateEnforcer = require('./middleware/state_enforcer.js')
 const retrieveCharge = require('./middleware/retrieve_charge.js')
 const resolveService = require('./middleware/resolve_service.js')
-
 const abTest = require('./utils/ab_test.js')
+
+const AB_TEST_THRESHOLD = process.env.AB_TEST_THRESHOLD || 100
 
 exports.paths = paths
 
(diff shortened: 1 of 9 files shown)


brid:
I've spoken with Jon about this and I think we need to have a level of confidence that any cookies this generates, don't risk affecting payment journeys, especially given we have multiple charges being attempted per session.

Knowing the lower/upper bounds on what affect this has on cookie sizes would get us most of the way there, preferably through monitoring this via Chrome developer tools or something similar.

tlwr:
This will add a cookie with a small key and a small value

It might cause some 502s but that is because of the chargeId cookie dropping not the AB testing cookie dropping

We should fix the chargeID cookie dropping, and merge this concurrently

app/router.js
```diff
@@ -52,7 +54,15 @@ exports.bind = function (app) {
   app.post(card.auth3dsHandler.path, middlewareStack, charge.auth3dsHandler)
   app.get(card.captureWaiting.path, middlewareStack, charge.captureWaiting)
   app.post(card.create.path, middlewareStack, charge.create)
-  app.get(card.confirm.path, middlewareStack, charge.confirm)
+  app.get(
+    card.confirm.path,
+    middlewareStack,
+    abTest.switch({
+      threshold: process.env.AB_TEST_THRESHOLD || 100,
```
maxcbc: As a general rule I would put this up top as a constant. 
```
const AB_TEST_THRESHOLD = process.env.AB_TEST_THRESHOLD || 100,
```

app/utils/ab_test.js
```diff
@@ -0,0 +1,80 @@
...
+
+exports.switch = function (opts) {
+  let threshold = Math.floor(opts.threshold || 90)
+  let defaultVariant = opts.defaultVariant
+  let testingVariant = opts.testingVariant
+
+  return function (req, res) {
+    let sessionValue = exports._getOrSetSession(req)
+
+    let showDefault = true
```
maxcbc: why is this explicitly true? why this variable?

app/utils/ab_test.js
```diff
@@ -0,0 +1,80 @@
...
+  if (exports._getSession(req) === null) {
+    logger.info('Could not find user ab testing session, creating one')
+    exports._setSession(req)
+  }
+  return exports._getSession(req)
+}
+
+exports.switch = function (opts) {
+  let threshold = Math.floor(opts.threshold || 90)
+  let defaultVariant = opts.defaultVariant
```
maxcbc: most of these can be `const` as they are never changed

app/utils/ab_test.js
```diff
@@ -0,0 +1,80 @@
...
+
+exports._setSession = function (req) {
+  return cookie.setSessionVariable(
+    req,
+    'abTestId',
+    exports.uniformlyGeneratedRandomNumber()
+  )
+}
+
+exports._getOrSetSession = function (req) {
```
maxcbc: slight style point, you can use fat-arrow `()=>{}` functions for all these.

app/utils/analytics.js
```diff
@@ -24,6 +24,7 @@ exports.withAnalytics = (charge, param) => {
       analyticsId: charge.gatewayAccount.analyticsId,
       type: charge.gatewayAccount.type,
       paymentProvider: charge.gatewayAccount.paymentProvider,
-      amount: charge.amount
+      amount: charge.amount,
+      testingVariant: 'original'
```
maxcbc: is this supposed to be hard coded?

tlwr: default value that is overwritten by the `_merge`

app/utils/ab_test.js
jonheslop: Is it just extra safe or is it silly? How should I refactor?

tlwr: The behaviour here is

- First calculate whether normally we would see the default
- Then if the showAbTest query param is truthy, force the ab test
- Then if the hideAbTest query param is truthy, force the default

This is superfluous, I guess we want a single query parameter to force/unforce

```
const showDefault = req.query.abTest
  ? req.query.abTest != 'yes'
  : sessionValue <= threshold
```

heathd:
This change does not significantly increase the likelihood of people hitting the cookie size limit. It adds one cookie value for the session overall, not per-charge. Currently I believe that you have to make of the order of 20 payments to exceed the cookie size limit (just did a quick test and 5 payments had a cookie size of 1499 bytes, 10 payments 2139 bytes). 

We should revisit the cookie size issue and see if we can get it to fail more gracefully rather than with a 500 error.

I don't think concerns about cookie size should prevent this being merged.

If others have reviewed the code I'm happy for this to be merged.

PR #392, opened 2018-04-18
PP-3579 - threshold needs to be an integer

jonheslop:


tlwr:
Yes

app/router.js
```diff
@@ -15,7 +15,7 @@ const retrieveCharge = require('./middleware/retrieve_charge.js')
 const resolveService = require('./middleware/resolve_service.js')
 const abTest = require('./utils/ab_test.js')
 
-const AB_TEST_THRESHOLD = process.env.AB_TEST_THRESHOLD || 100
+const AB_TEST_THRESHOLD = parseInt(process.env.AB_TEST_THRESHOLD, 10) || 100
```
georgeracu: Should we worry if `parseInt(process.env.AB_TEST_THRESHOLD, 10)` is > 100?

tlwr: No, the user's session value is never greater than 100, so if you set a threshold greater than 100 everyone still gets the default.

PR #400, opened 2018-05-01
PP-3703 base frontend on new govukpay node 8-x image

brid:
## WHAT
Implements frontend as being based on our govukpay/nodejs:8.11.1 Docker image.

Packages were updated to latest version, apart from 'change-case' which does funny things when going from 3-x to 4-x. This is a definite candidate for implementing our own version.

Some tests needed to be re-factored as nock request matching did not work in some of our tests, as things like supertest are adding their own headers to requests, and nock is more specific about comparing objects, and the order of properties.


test/models/charge_test.js
```diff
@@ -20,18 +20,18 @@ describe('updateStatus', function () {
     it('should return client unavailable', function () {
       var chargeModel = Charge('')
```
DanailMinchev: We still have "old style" `var`s

PR #402, opened 2018-05-02
PP-3673 - Check email for common typos

jonheslop:
Using the mailcheck NPM module we can check for common mistakes in
email domain names/TLDs.

If the email is technically valid it is then checked again by mailcheck
for typos, if one is flagged an error is shown with two radios the fixed
email and the origina, just in case mailcheck is wrong.

We then pass the choice and override the originally input email with the
choice.

So we can track how helpful this is, it also registers as an event to
Google Analytics so we can see if it's helping people. We only send the
domain to google i.e. gmail.com and not the first part so it keeps
everything anonymous.


package.json
```diff
@@ -48,9 +48,11 @@
     "csrf": "3.0.x",
     "express": "4.16.x",
     "forever": "0.15.x",
+    "gaap-analytics": "^3.0.0",
```
DanailMinchev: Probably need to regenerate `npm-shrinkwrap.json`

PR #431, opened 2018-06-20
PP-3905 Remove 'forever'

brid:
## WHAT
Removes the 'forever' module.

Also added missing ab_test tests which were failing NYC coverage thresholds


jonheslop:
https://bukk.it/omgyay.gif

PR #432, opened 2018-06-28
PP-3133: Route unknown pages on public facing apps to standard gov.uk/404

soniaturcotte:
## WHAT
Make 404 page work

## HOW 
Redirect frontend to standard gov.uk/404 page

PR #433, opened 2018-06-28
PP-3133: Route unknown pages on public facing apps to standard gov.uk/404

tlwr:
#432 raised via alphagov github so pr builder works

### DIFF
--- a/server.js
+++ b/server.js
@@ -166,7 +166,7 @@ function initialise () {
   initialiseCookies(app)
   initialiseTemplateEngine(app)
   initialisePublic(app)
-  initialiseRoutes(app)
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -165,8 +165,8 @@ function initialise () {
   initialisei18n(app)
   initialiseCookies(app)
   initialiseTemplateEngine(app)
-  initialiseRoutes(app)
   initialisePublic(app)
+  initialiseRoutes(app)
```
danworth: If it is imperative that this comes last it would be good to add a comment here to protect against future mishaps, as you've done for app/router.js.

PR #434, opened 2018-07-05
Added environment variables to permit running locally via pay local tool.

danworth:
Added environment variable ```USE_HTTP_CLIENT``` to use http client instead of https for local testing purposes. Added environment variable ```DISABLE_APPMETRICS``` to ignore appmetric import for running locally.

### DIFF
--- a/app/utils/base_client.js
+++ b/app/utils/base_client.js
@@ -7,11 +7,20 @@
  * Leaving for backward compatibility.
  */
 const urlParse = require('url')
-const https = require('https')
 const logger = require('winston')
+const https = setHttpClient()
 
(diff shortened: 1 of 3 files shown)


tlwr:
Yes but is consistent with self-service and products ui in terms of env var naming

mrlumbu:
We should make it consistent

danworth:
```DISABLE_APPMETRICS``` is consistent with the same products-ui change we made yesterday. I've changed ```USE_HTTP_CLIENT``` to ```DISABLE_INTERNAL_HTTPS``` to be consistent with existing variables. Also a corresponding change to the test.env so that ```CARDID_HOST``` now uses http.

PR #443, opened 2018-08-08
PP-4090 - reduce js bundle size

jonheslop:
After migrating to [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend) and looking at how it effected CSS/JS bundle size I noticed how large our JS was. This was mainly becausse of jQuery. So I have removed it as we weren’t using it too heavily anyway.

Also noticed in the bundle a lot of polyfills duplicated across the file. After a google I found this which confirmed by suspicions https://github.com/babel/babelify/issues/246

## With jQuery 😞 

File | Size
--- | ---
application.js | 654kB
application.min.js | 271kB

## Without jQuery 😑 

File | Size
--- | ---
application.js | 382kB
application.min.js | 184kB

## Refactored babel transform method 😄 

File | Size
--- | ---
application.js | 116kB
application.min.js | 51kB

(and a week ago, [we weren’t minifying our JS](https://github.com/alphagov/pay-frontend/pull/441) and it was 636kB 😱)


jonheslop:
image

Credit: @nickcolley

app/assets/javascripts/modules/form-validation.js
```diff
@@ -1,198 +1,249 @@
...
+  var countrySelect = document.getElementById('address-country')
+  var postcodeInput = document.getElementById('address-postcode')
+  var cardInput = document.getElementById('card-no')
+  var errorSummary = document.getElementsByClassName('govuk-error-summary')[0]
+  var logger = { info: function () { } }// replace with console to see output
+  // window.card comes from the view
+  var chargeValidations = module.chargeValidation(
+    i18n.chargeController.fieldErrors,
+    logger,
+    window.Card
```
nickcolley: In the previous code we have `$.extend({}, window.Card)` which i would assume is to send an object if window.Card is not defined or potentially it is cloning window.Card to avoid mutations to the original window.Card instance?

nickcolley:
Also be aware the JavaScript here may need polyfilling depending on what browsers you are supporting with this code.

app/assets/javascripts/modules/form-card-type.js
```diff
@@ -142,19 +142,24 @@ var showCardType = function() {
...
+      request.open('POST', '/check_card/' + chargeId, true);
+      request.setRequestHeader('Content-type', 'application/json');
+      request.onload = function() {
+        return resolve()
+      }
+      request.onerror = function() {
+        return reject({text: request.response.message})
+      }
+      request.send(JSON.stringify({
+        cardNo: cardInput.value.replace(/\D/g, '')
```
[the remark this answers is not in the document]
jonheslop: Good shout, it doesn’t send the request unless its entirely numeric, also we have a firewall watching requests and is has a whitelist of acceptable characters and it blocks if it receives something nasty

app/views/charge.njk
```diff
@@ -326,13 +326,13 @@
         </div>
       {% endif %}
```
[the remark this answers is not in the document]
jonheslop: Lol

app/assets/javascripts/modules/form-validation.js
jonheslop: Hmmm yeah, `window.Card` comes from here — https://github.com/alphagov/pay-frontend/blob/1029efd307e4ef4e5b225607cb6d55164cc2fcf0/app/views/includes/scripts.njk#L13-L17 so at worst it’s going to defined but empty. Which is unlikely, but does fail gracefully.

nickcolley: Are you sure the extension wasnt done to avoid mutating the original window.Card? Seems odd that someone would write this code otherwise.

PR #450, opened 2018-08-15
PP-4132 Add x-ray and continuation local storage to segment incoming/outgoing requests

brid:
## WHAT

Follows the patterns set in pay-selfservice to add x-ray tracing to pay-frontend, with a 1% sampling rate.

Automatic segment creation is handled by the aws-sdk for incoming requests, and outgoing requests to downstream services have been modified to include an x-ray trace header.

In order to get around the issues with segments being lost inside native promise chains, we have leveraged the 'continuation-local-storage' library to persist segment information through function chains without having to specifically pass segment information down through these.

In order to provide a specific example of sub-segmenting, we have added bespoke AWSXRay.captureAsyncFunc(..) calls to wrap existing functions related to the check_card flow, which happens in response to a credit card number being entered in the payment screens and then on focus exit or tabbing through the controls, an async call is made to check the card details.

The result of this is sub-segments can be created to isolate particular areas of interest, which for the sake of a current initiative is a specific area where telemetry is required.


app/middleware/retrieve_charge.js
```diff
@@ -1,26 +1,38 @@
...
     req.chargeId = chargeId
-    Charge(req.headers[CORRELATION_HEADER]).find(chargeId)
-      .then(data => {
-        req.chargeData = data
-        next()
-      })
-      .catch(() => {
-        views.display(res, 'SYSTEM_ERROR', withAnalyticsError())
-      })
+    AWSXRay.captureAsyncFunc('Charge_find', function (subsegment) {
```
mrlumbu: I think we need to be clear about where we introduce the subsegments, and how we name them. At the moment you have this subsegment initiated from a middleware, and the other one initiated from a model, but they are both named quite generically. I think it would make more sense to initiate all subsegments from the same layer, and probably the model layer is as good as any. So I therefore suggest this bit of instrumentation is moved to the `Charge.find` method

brid: So in terms of sub-seqments my rationale here was that we can nest these in order to get greater insight about a whole stack of function calls, with the 'check card' process being one of particular interest to the perf team at the time. We can of course do these all from the same layer but in doing that we might lose any granularity that may or may not be needed.

I agree though, the naming conventions and proper treatment of these would be great to iron out, given that this was essentially a POC implementation for a specific use case at the time.

app/router.js
```diff
@@ -21,7 +29,22 @@ const resolveService = require('./middleware/resolve_service.js')
 exports.paths = paths
 
 exports.bind = function (app) {
-  'use strict'
+  AWSXRay.enableManualMode()
+  AWSXRay.setLogger(logger)
+  AWSXRay.middleware.setSamplingRules('aws-xray.rules')
+  AWSXRay.config([AWSXRay.plugins.ECSPlugin])
+  app.use(AWSXRay.express.openSegment('pay_frontend'))
```
oswaldquek: do we need a `app.use(AWSXRay.express.closeSegment())` after the definition of routes? See "Tracing Incoming Requests with Express" at https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-nodejs-middleware.html

brid: We do, good spot thanks :-)

Sorry, this PR was not ready for prime time. I've updated the title as such.

oswaldquek: sorry, i guess i'm eager to get this merged as it will help will perf testing ;) no worries though, i'm off tomorrow and am occupied with something else.

app/middleware/retrieve_charge.js
mrlumbu: Having better understood how these segments are composed, I agree that this is fine for now

app/utils/base_client.js
```diff
@@ -38,6 +45,20 @@ if (process.env.DISABLE_INTERNAL_HTTPS !== 'true') {
...
 const agent = new https.Agent(agentOptions)
 
+const getHeaders = function getHeaders (args, segmentData) {
+  let headers = {}
+  headers['Content-Type'] = 'application/json'
+  headers[CORRELATION_HEADER_NAME] = args.correlationId || ''
+
+  if (segmentData.clsSegment) {
+    const subSegment = segmentData.subSegment || new AWSXRay.Segment('_request', null, segmentData.clsSegment.trace_id)
+    headers['X-Amzn-Trace-Id'] = 'Root=' + segmentData.clsSegment.trace_id + ';Parent=' + subSegment.id + ';Sampled=1'
```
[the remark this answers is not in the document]
brid: Love it! Adding that.

PR #453, opened 2018-08-20
PP-4097 reconfigure app to use i18n dynamically

jonheslop:
Reconfigure app to use i18n in a way which means english isn’t hard coded.

This is also changes the way the strings are fed into the the templates which required switching to using `_("object.path")` syntax

Also added cy.json which contains the welsh strings

### DIFF
--- a/app/middleware/resolve_language.js
+++ b/app/middleware/resolve_language.js
@@ -1,0 +1,9 @@
+'use strict'
+
+// local dependencies
+const i18n = require('i18n')
+
+module.exports = function (req, res, next) {
+  res.locals.translationStrings = JSON.stringify(i18n.getCatalog('en'))
+  next()
+}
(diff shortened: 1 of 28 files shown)


jonheslop:
Hi Alex,

I think all of this is PP-4097, the tasks of the ticket are

- Move all hardcoded English text out of templates and into en.json
- Create cy.json file
- Ensure en.json and cy.json have the same keys

That’s what this ticket does. It’s not concerned with any of the tasks in PP-4098

alexbishop1:
Jon and I have discussed this in person and he explained to me that the code changes are just internationalisation changes to make supporting multiple languages easier but do not include any of the work to switch based on the language of the payment (though there is a facility to switch based on a URL parameter to demonstrate it works).

Since it’s a bit ambiguous whether such internationalisation changes should be part of this story or PP-4098, I’m happy for them to be done here. These changes will pave the way for PP-4098, which will focus on how to determine which language to use, rather than mechanics of actually supporting multiple languages.

Jon has a agreed to move the addition of the `cy.json` file into another PR.

PR #455, opened 2018-08-20
bau: log memory stats

oswaldquek:
As part of the performance testing epic
(https://payments-platform.atlassian.net/browse/PP-4004) we have made some
changes to improve performance Some of these are:

- caching token requests in publicapi
- upgrading frontend's ec2 instances from m3.medium -> c4.large

It would be good to log the memory statistics for each server process in node
should we need to further improve performance (for example, by changing the
heap size, which would then affect when the garbage collection kicks in) for
frontend in the future. We currently have no easy way to know this.

@oswaldquek

### DIFF
--- a/server.js
+++ b/server.js
@@ -6,6 +6,7 @@ if (!process.env.DISABLE_APPMETRICS) {
 // Node.js core dependencies
 const path = require('path')
 const fs = require('fs')
+const v8 = require('v8')
 
 // npm dependencies
 const express = require('express')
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -77,6 +78,9 @@ function initialiseGlobalMiddleware (app) {
   app.use(compression())
 
   app.disable('x-powered-by')
+
+  logger.info('Heap statistics: ' + v8.getHeapStatistics())
```
brid: This code will only execute once when the app starts up. Is that the intention?

oswaldquek: i think for now yes. @heathd was this what you were suggesting this morning, that we log memory stats when the app starts up? `getHeapStatistics` will give us the total heap size (see https://nodejs.org/api/v8.html#v8_v8_getheapstatistics)

PR #467, opened 2018-09-06
PP-4098 - CardID messages need to be translatable too

jonheslop:
Using i18n to get strings from appropriate JSON file for error messages from CardID

Also put the i18n config in it’s own file so we can use in app setup and controller config


locales/cy.json
```diff
@@ -96,7 +96,10 @@
 				"numberIncorrectLength": "Nid yw rhif y cerdyn yr hyd cywir",
 				"luhnInvalid": "Rhowch rif cerdyn dilys",
 				"cardNotSupported": "Nid yw’r math hwn o gerdyn yn cael ei dderbyn",
-				"nonNumeric": "Rhowch rif cerdyn dilys"
+				"nonNumeric": "Rhowch rif cerdyn dilys",
```
DanailMinchev: I guess the translation is correct :)

PR #472, opened 2018-09-14
[WIP] PP-4183 Change cipher

mrlumbu:
We have reason to believe that using a less rigorous cipher for
tls may speed up node http calls. Let's see


whpearson:
Closing as we have off-loaded tls to nginx.

PR #477, opened 2018-09-20
PP-4213: Set the host header and redirect to forward proxy

oswaldquek:
As we will be sending requests to the nginx forward proxy, nginx will need
information about where to forward the request to. This will be got from the
host header.

@oswaldquek

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -46,7 +46,7 @@ pipeline {
                 }
             }
             steps {
-                runCardPaymentsE2E("frontend")
+                runCardPaymentsE2E("frontend", null, "PP-4213")
             }
         }
          stage('ZAP Tests') {
--- a/app/utils/add_proxy.js
+++ b/app/utils/add_proxy.js
@@ -1,0 +1,24 @@
+'use strict'
+
+//npm dependencies
+const urlParse = require('url').parse
+const _ = require('lodash')
+
+// constants
+const {FORWARD_PROXY_URL} = process.env
+
+const _addProxy = function addProxy (url) {
+  const parsedUrl = urlParse(url)
+
+  if (FORWARD_PROXY_URL) {
+    const parsedProxyUrl = urlParse(FORWARD_PROXY_URL)
+    return _.merge(parsedUrl, _.pick(parsedProxyUrl, ['hostname', 'port']))
+  }
+  return parsedUrl
+}
+
+module.exports = {
+  addProxy: function (url) {
+    return _addProxy(url)
+  }
+}
--- a/app/utils/add_proxy.js
+++ b/app/utils/add_proxy.js
@@ -1,6 +1,6 @@
 'use strict'
 
-//npm dependencies
+// npm dependencies
 const urlParse = require('url').parse
 const _ = require('lodash')
 
--- a/app/utils/add_proxy.js
+++ b/app/utils/add_proxy.js
@@ -7,7 +7,7 @@ const _ = require('lodash')
 // constants
 const {FORWARD_PROXY_URL} = process.env
 
-const _addProxy = function addProxy (url) {
+module.exports.addProxy = function addProxy (url) {
   const parsedUrl = urlParse(url)
 
   if (FORWARD_PROXY_URL) {
@@ -15,10 +15,4 @@ const _addProxy = function addProxy (url) {
     return _.merge(parsedUrl, _.pick(parsedProxyUrl, ['hostname', 'port']))
   }
   return parsedUrl
-}
-
-module.exports = {
-  addProxy: function (url) {
-    return _addProxy(url)
-  }
 }
--- a/app/utils/base_client.js
+++ b/app/utils/base_client.js
@@ -17,6 +17,7 @@ const AWSXRay = require('aws-xray-sdk')
 // Local dependencies
 const customCertificate = require('./custom_certificate')
 const CORRELATION_HEADER_NAME = require('./correlation_header').CORRELATION_HEADER
+const {addProxy} = require('./add_proxy')
 
 const agentOptions = {
   keepAlive: true,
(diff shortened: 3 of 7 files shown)


app/utils/base_client.js
```diff
@@ -79,18 +75,27 @@ const getHeaders = function getHeaders (args, segmentData) {
 const _request = function request (methodName, url, args, callback, subSegment) {
   const namespace = getNamespace(clsXrayConfig.nameSpaceName)
   const clsSegment = namespace ? namespace.get(clsXrayConfig.segmentKeyName) : null
-  const parsedUrl = urlParse.parse(url)
+  const parsedUrl = urlParse(url)
+
+  var urlOrForwardProxy
```
mrlumbu: rather than doing it like this I would prefer to encapsulate this logic in a function, so 
```
const addProxy = function addProxy (url) {
  const parsedUrl = urlParse(url)

  if (process.env.FORWARD_PROXY_URL) {
    const parsedProxyUrl = urlParse(process.env.FORWARD_PROXY_URL)
    return _.merge(parsedUrl, _.pick(parsedProxyUrl, ['hostname', 'port']))
  }
  return parsedUrl
}
```

then

```
const proxiedUrl = addProxy(url)
  const httpsOptions = {
    hostname: proxiedUrl.hostname,
    port: proxiedUrl.port,
    path: proxiedUrl.pathname,
...
```

app/utils/base_client.js
```diff
@@ -79,18 +75,27 @@ const getHeaders = function getHeaders (args, segmentData) {
...
+  if (process.env.FORWARD_PROXY_URL) {
+    urlOrForwardProxy = parsedUrl
+    urlOrForwardProxy.hostname = urlParse(process.env.FORWARD_PROXY_URL).hostname
+    urlOrForwardProxy.port = urlParse(process.env.FORWARD_PROXY_URL).port
+  } else {
+    urlOrForwardProxy = parsedUrl
+  }
 
   const httpsOptions = {
```
mrlumbu: `httpOptions`

app/utils/base_client2.js
```diff
@@ -79,11 +84,22 @@ const _request = function request (methodName, url, args, callback, subSegment)
   const namespace = getNamespace(clsXrayConfig.nameSpaceName)
   const clsSegment = namespace ? namespace.get(clsXrayConfig.segmentKeyName) : null
 
+  var urlOrForwardProxy
```
mrlumbu: same as for base_client. In fact, if it is identical, maybe make it a utility and then you could even write a test or two for it, which is probably a good idea

Jenkinsfile
```diff
@@ -46,7 +46,7 @@ pipeline {
                 }
             }
             steps {
-                runCardPaymentsE2E("frontend")
+                runCardPaymentsE2E("frontend", null, "PP-4213")
```
mrlumbu: you'll need to revert this, and do some sort of backwards compatibilty dance, or leave it in, but make sure you merge the relevant other branch and clean this up immediately after

app/utils/base_client.js
```diff
@@ -41,14 +32,19 @@ if (process.env.DISABLE_INTERNAL_HTTPS !== 'true') {
...
-const getHeaders = function getHeaders (args, segmentData) {
+const getHeaders = function getHeaders (args, segmentData, url) {
   let headers = {}
   headers['Content-Type'] = 'application/json'
   headers[CORRELATION_HEADER_NAME] = args.correlationId || ''
+  if (url) {
+    var port = (urlParse(url).port) ? ':' + urlParse(url).port : ''
+    headers['host'] = urlParse(url).hostname + port
+  }
+  logger.debug('headers: ' + JSON.stringify(headers))
```
mrlumbu: probs remove this

Jenkinsfile
oswaldquek: yup, this pay-frontend branch will need to be merged first, before https://github.com/alphagov/pay-scripts/pull/707 can be promptly merged.

oswaldquek: then once https://github.com/alphagov/pay-scripts/pull/707 is merged, I can remove the "PP-4213" here.

app/utils/add_proxy.js
```diff
@@ -0,0 +1,24 @@
+'use strict'
+
+// npm dependencies
+const urlParse = require('url').parse
+const _ = require('lodash')
+
+// constants
+const {FORWARD_PROXY_URL} = process.env
+
+const _addProxy = function addProxy (url) {
```
mrlumbu: I would change this to 
```
module.exports.addProxy = function addProxy (url) {
...
```
then remove the bottom exports bit

test/utils/add_proxy_test.js
```diff
@@ -0,0 +1,31 @@
...
+const {expect} = require('chai')
+const urlParse = require('url').parse
+const proxyquire = require('proxyquire').noPreserveCache()
+
+describe('add proxy test', () => {
+  describe('with FORWARD_PROXY_URL env var', () => {
+    it('url should be proxied with new hostname and port', () => {
+      const forwardProxy = urlParse('http://forward-proxy:8888')
+      process.env.FORWARD_PROXY_URL = forwardProxy.href
+      const addProxy = proxyquire('../../app/utils/add_proxy', {})
```
mrlumbu: don't think this needs to be proxyquired - you aren't stubbing anything

oswaldquek: unfortunately it doesn't work if i don't use proxyquired. this test doesn't pass:
```
const {addProxy} = require('../../app/utils/add_proxy')

describe('add proxy test', () => {
  describe('with FORWARD_PROXY_URL env var', () => {
    it('url should be proxied with new hostname and port', () => {
      const forwardProxy = urlParse('http://forward-proxy:8888')
      process.env.FORWARD_PROXY_URL = forwardProxy.href
      const proxiedUrl = addProxy(urlParse('http://localhost:8080'))
      expect(proxiedUrl.hostname).to.equal(forwardProxy.hostname)
      expect(proxiedUrl.port).to.equal(forwardProxy.port)
    })
  })
```
i actually got the idea to use proxyquire from an ex colleague's blog ;) https://medium.com/@maxcbc/mocking-environment-variables-in-node-js-a17a416e127c

PR #493, opened 2018-10-04
PP-4253 payment request api

jonheslop:
If a gateway is configured to accept native webpayments using [this boolean](https://github.com/alphagov/pay-connector/pull/722) then we show an option to the user to use Payment Request API, depending on the browser/device, this could be Apple Pay or Google/Samsung/Android/Windows Pay

screen shot 2018-10-10 at 11 23 17

## Generic web payments flow

The user confirms the payment in the native dialogue and then their card details are passed to our controller and we create and capture the charge all in one, skipping the confirm payment page completely.

web-payments-api
_it redirects to the homepage, but IRL that would redirect to a payment success page_

## Apple Pay Flow
If it’s Apple Pay we have to first validate the session by dialing Apple to get confirmation we’re a registered merchant, once this is done the payment authorisation can be made on the users, phone or watch.

apple-pay


app/assets/javascripts/browsered/payment-request.js
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    session.onpaymentauthorized = event => {
+      // Send payment for processing...
+      const payment = event.payment;
+      console.log('authorisation complete', payment)
+
+      // ...return a status and redirect to a confirmation page
+      session.completePayment(ApplePaySession.STATUS_SUCCESS);
+      // window.location.href = "/success.html";
```
brid: Should this be here?

brid: Double checked with Jon and this gets stripped out in production when it's minified.

PR #505, opened 2018-10-15
PP-4196 Corporate card information on confirmation page

DanailMinchev:
## WHAT

- Added `charge.corporate_surcharge` which is corporate card fee and can be present or not to the `confirm.njk` template.
- Added `charge.total_amount` which is total amount of the charge and can be present or not to the `confirm.njk` template.
- Added UI test with connector mock.

with @stephencdaly

### DIFF
--- a/app/views/confirm.njk
+++ b/app/views/confirm.njk
@@ -69,14 +69,20 @@
     </form>
   </div>
   <div class="govuk-grid-column-one-third payment-summary-wrap">
-    <aside class="payment-summary">
+    <aside class="payment-summary govuk-!-font-size-19">
       <h2 class="govuk-heading-m">{{ __("paymentSummary.title") }}</h2>
-      <p id="payment-description" class="govuk-body govuk-!-font-weight-bold govuk-!-font-size-39">
+      <p id="payment-description" class="govuk-body govuk-!-font-weight-bold">
         {{ charge.description }}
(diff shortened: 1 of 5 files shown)


app/views/confirm.njk
```diff
@@ -69,14 +69,20 @@
     </form>
   </div>
   <div class="govuk-grid-column-one-third payment-summary-wrap">
-    <aside class="payment-summary">
+    <aside class="payment-summary govuk-!-font-size-19">
       <h2 class="govuk-heading-m">{{ __("paymentSummary.title") }}</h2>
-      <p id="payment-description" class="govuk-body govuk-!-font-weight-bold govuk-!-font-size-39">
```
jonheslop: I dont know how `govuk-!-font-size-39` ever got here (although I can see in the history it was me 😬 ) `govuk-body` sets the size to 19px so it‘s redundant.

app/views/confirm.njk
```diff
@@ -69,14 +69,20 @@
     </form>
   </div>
   <div class="govuk-grid-column-one-third payment-summary-wrap">
-    <aside class="payment-summary">
+    <aside class="payment-summary govuk-!-font-size-19">
```
jonheslop: Remove the font size override, it’s not doing anything

PR #506, opened 2018-10-16
PP-4263 Point models and controllers at new base client

brid:
- Refactor code around Promises rather than callbacks
- Fix tests
- A base client test relating to asserting keep-alive connections was removed, as the new client/pattern does not let us do the same determinations as last time


test/test.env
```diff
@@ -5,4 +5,4 @@ COOKIE_MAX_AGE=5400000
 SESSION_ENCRYPTION_KEY=naskjwefvwei72rjkwfmjwfi72rfkjwefmjwefiuwefjkbwfiu24fmjbwfk
 DISABLE_INTERNAL_HTTPS=true
 CSRF_USER_SECRET=123456789012345678
-CARDID_HOST=http://cardid.pymnt.localdomain
+CARDID_HOST=http://cardid.pymnt.localdomain:65530
```
jonheslop: Is this ok?

brid: Yep, the forward proxy in a test scenario requires a port be available.

PR #507, opened 2018-10-16
PP-4295 ie client side js fixes

jonheslop:
- IE can’t loop on NodeLists so convert to standard array
- IE can’t do native promises so included a polyfill


app/views/includes/scripts.njk
```diff
@@ -1,7 +1,9 @@
-<!-- Javascript -->
+{# Javascript #}
 <script type="text/javascript">
   var i18n = {{ translationStrings | safe }};
 </script>
+{# Promise polyfill for IE - https://github.com/taylorhakes/promise-polyfill#cdn-polyfill-use #}
+<script src="https://cdn.jsdelivr.net/npm/promise-polyfill@8/dist/polyfill.min.js"></script>
```
DanailMinchev: I think it's better to download this dependency and host it ourselves in `assets/` directory.
Also, as discussed with @brid we can add `integrity` parameter to check if the file was altered somehow.

PR #510, opened 2018-10-23
PP-4310 - removed links to nothing from header

jonheslop:
For standard GOV.UK Pay payment pages this means the service name
doesn’t link to # it just isnt a link anymore.

For custom branding pages neither the logo or the service names are
links anymore as we wouldn’t know where to point them

Had to add a new selector to make sure the text is in the right colour, as the old selector was looking for a link specifically


jonheslop:
Ha thanks, it was there all along, but was coming from the Nunjucks macro, but we can’t use that because it always makes a link when we dont want one 😞

PR #514, opened 2018-10-30
PP-4205 Use direct redirect property

stephencdaly:
## WHAT
- Use new 'redirect_to_service_immediately_on_terminal_state' property returned by adminusers GET service call when resolving the service in middleware.
- Modified service test fixture to expect new property.
- Made some minor alterations to some tests:
  - The happy path test in resolve_service_test now asserts on something that will prove the resolve was successful.
  - Use the service fixture to construct the input JSON in service_test.

with @DanailMinchev

### DIFF
--- a/test/fixtures/service_fixtures.js
+++ b/test/fixtures/service_fixtures.js
@@ -1,17 +1,16 @@
 'use strict'
 
-// Custom dependencies
+// Local dependencies
 const pactBase = require('./pact_base')
+const random = require('../../app/utils/random')
 
 // Global setup
 const pactServices = pactBase({array: ['service_ids']})
 
-let random = require('../../app/utils/random')
-
 module.exports = {
   validServiceResponse: (serviceData = {}) => {
-    let defaultCustomBranding = {css_url: 'css url', image_url: 'image url'}
-    let defaultMerchantDetails = {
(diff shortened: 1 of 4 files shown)


test/fixtures/service_fixtures.js
```diff
@@ -11,7 +11,7 @@ let random = require('../../app/utils/random')
 module.exports = {
   validServiceResponse: (serviceData = {}) => {
     let defaultCustomBranding = {css_url: 'css url', image_url: 'image url'}
-    let defaultMerchantDetais = {
+    let defaultMerchantDetails = {
```
jonheslop: Maybe don’t worry about it now but Mark and I try to factor out all uses of `var` and `let` (unless the variable value is modified, then `let` is fine) in files.

It impacts peformace, but this is a test I guess

test/models/service_test.js
```diff
@@ -1,14 +1,16 @@
+'use strict'
```
jonheslop: Nice

test/fixtures/service_fixtures.js
stephencdaly: We've done it anyway, added another commit to this PR for this

PR #519, opened 2018-11-01
PP-4205 Implement redirect to service

stephencdaly:
## WHAT
- Move the `redirectToServiceImmediatelyOnTerminalState` property on the `Service` model to correct place
- Perform a redirect to the return URL when a terminal state in the payment flow is reached, rather than show our pages

## HOW 
- Modified `views.js` to add an additional `terminalView` property on action objects to determine whether a view is a terminal state
- In `views.display`, if the action has the `terminalView` set to `true` and `redirect_to_service_immediately_on_terminal_state` is set to `true` on the `Service`, perform a redirect to the return URL
- Added tests to test the new logic introduced to `views.display`
- Originally we were attempting to cancel the charge before the redirect if an error occurred leaving it in a cancellable state. However after discussion we reverted this change, as it didn't make sense to change the charge status to `USER CANCELLED` in this case. However, we modified some code in `return_controller.js` around cancellation to allow for this, which we left in as we thought our changes were an improvement.

with @DanailMinchev

### DIFF
--- a/app/controllers/charge_controller.js
+++ b/app/controllers/charge_controller.js
@@ -8,18 +8,18 @@ const {getNamespace} = require('continuation-local-storage')
 const AWSXRay = require('aws-xray-sdk')
 
 // local dependencies
-const logging = require('../utils/logging.js')
-const views = require('../utils/views.js')
-const normalise = require('../services/normalise_charge.js')
-const chargeValidator = require('../utils/charge_validation_backend.js')
-const Charge = require('../models/charge.js')
-const Card = require('../models/card.js')
-const State = require('../models/state.js')
-const paths = require('../paths.js')
-const CORRELATION_HEADER = require('../utils/correlation_header.js').CORRELATION_HEADER
-const {countries} = require('../services/countries.js')
-const {commonTypos} = require('../utils/email_tools.js')
-const {withAnalyticsError, withAnalytics} = require('../utils/analytics.js')
+const logging = require('../utils/logging')
+const responseRouter = require('../utils/response_router')
+const normalise = require('../services/normalise_charge')
+const chargeValidator = require('../utils/charge_validation_backend')
+const Charge = require('../models/charge')
+const Card = require('../models/card')
+const State = require('../models/state')
+const paths = require('../paths')
+const CORRELATION_HEADER = require('../utils/correlation_header').CORRELATION_HEADER
+const {countries} = require('../services/countries')
+const {commonTypos} = require('../utils/email_tools')
+const {withAnalyticsError, withAnalytics} = require('../utils/analytics')
 const connectorClient = require('../services/clients/connector_client')
 
 // constants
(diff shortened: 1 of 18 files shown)


mrlumbu:
Looking at the commits it seems like you went down a certain path, then rolled back to something previously. Could you squash stuff to make the commits tell a clean story of how you got to where you are, rather than all the stuff along the way.

app/utils/views.js
```diff
@@ -205,20 +226,27 @@ module.exports = {
...
     options.viewName = viewName
     if (!action) {
       logger.error('VIEW ' + viewName + ' NOT FOUND')
       options = {viewName: 'error'}
       action = this.ERROR
     }
+    const redirectToServiceImmediatelyOnTerminalState =
+      res.locals &&
+      res.locals.service &&
+      res.locals.service.redirectToServiceImmediatelyOnTerminalState
```
mrlumbu: this can be `_.get(res, 'locals.service. redirectToServiceImmediatelyOnTerminalState', false)`

app/utils/views.js
```diff
@@ -205,20 +226,27 @@ module.exports = {
...
+    if (action.terminalView && redirectToServiceImmediatelyOnTerminalState && req.chargeData) {
```
mrlumbu: feels like these are other conditions that could just be added to the boolean `redirectToServiceImmediatelyOnTerminalState`

app/utils/views.js
```diff
@@ -205,20 +226,27 @@ module.exports = {
...
-    let action = lodash.result(this, viewName)
     options = options || {}
+    let action = lodash.result(this, viewName)
     options.viewName = viewName
     if (!action) {
       logger.error('VIEW ' + viewName + ' NOT FOUND')
       options = {viewName: 'error'}
       action = this.ERROR
     }
+    const redirectToServiceImmediatelyOnTerminalState =
```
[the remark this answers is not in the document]
DanailMinchev: As discussed with @mrlumbu and @stephencdaly we will stick with this implementation, but we will change the name of the file as well as the name of the method, because we want to have central place to control the redirection.

app/controllers/charge_controller.js
```diff
@@ -254,21 +254,21 @@ module.exports = {
                 break
               case 500:
                 logging.failedChargePost(409)
```
mrlumbu: think this line and similar ones below are missing `req` as an argument? Unless I'm missing something

app/utils/response_router.js
```diff
@@ -205,24 +226,34 @@ module.exports = {
...
+    action = actions.ERROR
+  }
 
-    if (lodash.get(options, 'analytics.path')) {
-      options.analytics.path = options.analytics.path + lodash.get(action, 'analyticsPage', '')
-    }
-    res.status(action.code || 200)
-    res.render(action.view, options)
+  const directRedirectEnabled = lodash.get(res, 'locals.service.redirectToServiceImmediatelyOnTerminalState', false)
+  if (action.terminal && directRedirectEnabled && req.chargeData) {
```
mrlumbu: I would be tempted to roll all of this into a function
```
shouldRedirect = (req, res, action) => {
    const directRedirectEnabled = lodash.get(res, 'locals.service.redirectToServiceImmediatelyOnTerminalState', false)
    return (action.terminal && directRedirectEnabled && req.chargeData)
}
```

But no biggie - maybe do in a separate PR if you agree

app/controllers/charge_controller.js
DanailMinchev: @mrlumbu you are absolutely right! Did some bad rebasing and didn't push the changes :(

PR #520, opened 2018-11-02
PP-4347 Implement a Connector client on pay-frontend

brid:
Removes direct calls to connector from the charge controller, and instead implements a connector client that abstracts these calls.

### DIFF
--- a/app/controllers/charge_controller.js
+++ b/app/controllers/charge_controller.js
@@ -9,7 +9,6 @@ const AWSXRay = require('aws-xray-sdk')
 
 // local dependencies
 const logging = require('../utils/logging.js')
-const baseClient = require('../utils/base_client')
 const views = require('../utils/views.js')
 const normalise = require('../services/normalise_charge.js')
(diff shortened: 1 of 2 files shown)


app/controllers/charge_controller.js
```diff
@@ -238,62 +221,55 @@ module.exports = {
...
-              views.display(res, 'ERROR', withAnalytics(charge))
-            })
-          })
-          .catch(err => {
+    AWSXRay.captureAsyncFunc('Charge_patch', function (subSegment) {
+      Charge(req.headers[CORRELATION_HEADER]).patch(req.chargeId, 'replace', 'email', req.body.payerEmail, subSegment)
+        .then(() => {
+          subSegment.close()
+          const correlationId = req.headers[CORRELATION_HEADER] || ''
+          const payload = normalise.apiPayload(_.merge(req, convertedPayload), 'visa')
```
jonheslop: Aaah my bad, look when I was prototyping with PaymentRequest stuff, I hard coded `visa` 😬. Could leave for now perhaps, as  no one is using it anyway

PR #522, opened 2018-11-05
BAU - bump product page version to include titles fix

jonheslop:


jonheslop:
This was a duplicate of #517 to see if that would stop Jenkins being stupid, it didnt work

PR #525, opened 2018-11-06
PP-4347 charge controller refactor

brid:
A refactor of charge controller with the aim of reducing the sheer amount of code in the module file, moving some logic into more easily digestible chunks, and giving us more scope to further improve things as we go on.


app/controllers/return_controller.js
```diff
@@ -6,7 +6,7 @@ const logger = require('winston')
 // local dependencies
 const Charge = require('../models/charge')
 const responseRouter = require('../utils/response_router')
-const CORRELATION_HEADER = require('../utils/correlation_header').CORRELATION_HEADER
+const CORRELATION_HEADER = require('../../config/correlation_header').CORRELATION_HEADER
```
jonheslop: You could use destructuring here

jonheslop: and on all the others

PR #532, opened 2018-11-16
PP-4373 Separate 3DS controller

mrlumbu:
Separates out 3DS controller from charge controller, as its
a fairly self contained unit. I have not made any other changes
other than lifting and shifting, and amending the necessary routes.

### DIFF
--- a/app/controllers/three_d_secure_controller.js
+++ b/app/controllers/three_d_secure_controller.js
@@ -1,0 +1,121 @@
+
+// npm dependencies
+const _ = require('lodash')
+
+// local dependencies
+const responseRouter = require('../utils/response_router')
+const normalise = require('../services/normalise_charge')
+const paths = require('../paths')
+const {withAnalytics} = require('../utils/analytics')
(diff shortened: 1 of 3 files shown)


app/controllers/three_d_secure_controller.js
```diff
@@ -0,0 +1,121 @@
...
+const {withAnalytics} = require('../utils/analytics')
+const connectorClient = require('../services/clients/connector_client')
+
+// constants
+const {views, threeDsEPDQResults} = require('../../config/charge_controller')
+const {CORRELATION_HEADER} = require('../../config/correlation_header')
+
+const routeFor = (resource, chargeId) => paths.generateRoute(`card.${resource}`, {chargeId: chargeId})
+
+const redirect = res => {
```
simad: I think we now have this redirect object in multiple places - we could extract it as a utility method. It does not have to be done as part of this PR, though.

mrlumbu: yep 👍  plan to do that and similar for other methods in another PR

PR #544, opened 2018-11-23
PP-4394 Cookie deletion after capture/cancel

brid:
## WHAT
Deletes session[key] for all valid sessions, based on existence of encryption key, and the existence of relevant cookie on the request

with @SandorArpa

### DIFF
--- a/app/controllers/return_controller.js
+++ b/app/controllers/return_controller.js
@@ -6,12 +6,17 @@ const logger = require('winston')
 // Local dependencies
 const Charge = require('../models/charge')
 const responseRouter = require('../utils/response_router')
-const CORRELATION_HEADER = require('../../config/correlation_header').CORRELATION_HEADER
-const withAnalyticsError = require('../utils/analytics').withAnalyticsError
+const { CORRELATION_HEADER } = require('../../config/correlation_header')
+const { withAnalyticsError } = require('../utils/analytics')
+const cookies = require('../utils/cookies')
+const { createChargeIdSessionKey } = require('../utils/session')
 
 exports.return = (req, res) => {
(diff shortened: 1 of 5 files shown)


app/controllers/return_controller.js
```diff
@@ -6,12 +6,16 @@ const logger = require('winston')
...
-const withAnalyticsError = require('../utils/analytics').withAnalyticsError
+const { CORRELATION_HEADER } = require('../../config/correlation_header')
+const { withAnalyticsError } = require('../utils/analytics')
+const cookies = require('../utils/cookies')
 
 exports.return = (req, res) => {
   const correlationId = req.headers[CORRELATION_HEADER] || ''
   const charge = Charge(correlationId)
+  // Remove the charge data from the cookie
+  const cookieKey = `ch_${req.chargeId}`
```
stephencdaly: should this use session#createChargeIdSessionKey to make sure this doesn't get out of sync if how we construct this key changes?

app/utils/cookies.js
```diff
@@ -100,6 +112,22 @@ function getSessionCookieName () {
...
+ */
+function deleteSessionVariable (req, key) {
+  if (SESSION_ENCRYPTION_KEY) {
+    deleteValueOnCookie(req, key, SESSION_COOKIE_NAME_1)
+  }
+  if (SESSION_ENCRYPTION_KEY_2) {
+    deleteValueOnCookie(req, key, SESSION_COOKIE_NAME_2)
+  }
+}
+
```
stephencdaly: should we add some tests for this?

app/controllers/return_controller.js
sfount: 👍

PR #556, opened 2018-12-11
PP-4515 - preserve fields on reload

jonheslop:
If you have JS disabled and make an error you have to re-enter the four
card fields again. This is annoying and I am not sure if it’s necessary.

### DIFF
--- a/config/charge_controller.js
+++ b/config/charge_controller.js
@@ -14,5 +14,5 @@ module.exports = {
     declined: 'DECLINED',
     error: 'ERROR'
   },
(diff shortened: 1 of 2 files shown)


config/charge_controller.js
```diff
@@ -13,6 +13,5 @@ module.exports = {
     success: 'AUTHORISED',
     declined: 'DECLINED',
     error: 'ERROR'
-  },
-  preserveProperties: ['cardholderName', 'addressLine1', 'addressLine2', 'addressCity', 'addressPostcode', 'addressCountry', 'email']
```
sfount: It might be a good idea to keep the `preserveProperties` method of picking fields. This would just stop future sensitive fields from being merged implicitly.

jonheslop: Good point

PR #562, opened 2018-12-20
PP-4563 - Updating packages that depended on cryptiles

jonheslop:
Because of https://www.npmjs.com/advisories/720

interestingly both modules have just dropped Cryptiles rather than
update to non vulnerable version


jonheslop:
This got fixed hours before my first day back after Christmas ☺️ 
https://github.com/pact-foundation/pact-node/issues/132#issuecomment-451042542

PR #566, opened 2019-01-16
[WIP] PP-4365 Apple Pay hooking things together

jonheslop:
We can invoke a payment client side and we can process one in [pay-connector](https://github.com/alphagov/pay-connector/) this PR joins the two. We’re struggling to test locally because you can’t authorise a domain with Apple and stubbing for tests is tricky too. So just cobbled it together and are going to see if it works on test. If it does we can confidently write tests and stubs


app/controllers/apple-pay/handle-auth-response-controller.js
```diff
@@ -0,0 +1,81 @@
...
+    toAuthWaiting: (chargeId) => res.redirect(303, routeFor('authWaiting', chargeId)),
+    toConfirm: (chargeId) => res.redirect(303, routeFor('confirm', chargeId)),
+    toNew: (chargeId) => res.redirect(303, routeFor('new', chargeId)),
+    toReturn: (chargeId) => res.redirect(303, routeFor('return', chargeId))
+  }
+}
+
+const handleAuthResponse = (req, res, charge) => response => {
+  switch (response.statusCode) {
+    case 202:
```
sfount: Comments for how these cases should be handled in the future or just removing this cases would be useful for future devs.

jonheslop:
Thinking should leave the merge of this until Monday, as I am out tomorrow and mush time left to today

PR #568, opened 2019-01-17
BAU log request start at info level

heathd:
it seems useful to log this as well as request completion. It can help
debugging in cases where something goes wrong before the response is received.

### DIFF
--- a/app/utils/request_logger.js
+++ b/app/utils/request_logger.js
@@ -4,7 +4,7 @@
 const logger = require('winston')
 
 exports.logRequestStart = context => {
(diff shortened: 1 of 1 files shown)


mocha-test-results.txt
```diff
@@ -0,0 +1,574 @@
+warn: DISABLE_INTERNAL_HTTPS is set.
```
sfount: Is this results file checked in intentionally?

heathd: no. doh!

PR #570, opened 2019-01-17
PP-4651 Improve testing

georgeracu:
## WHAT
- Added functional test for state enforcer when payment provider is
Stripe. Stripe charges with status AUTH_3DS_READY from connector should
go to auth_waiting page
- Improved logging to have the payment provider and charge id in case
other charges fail from the same reason
- Updated code syntax

with @kbottla

### DIFF
--- a/test/middleware/state_enforcer_test.js
+++ b/test/middleware/state_enforcer_test.js
@@ -1,18 +1,20 @@
-var path = require('path')
-var assert = require('assert')
-var expect = require('chai').expect
-var stateEnforcer = require(path.join(__dirname, '/../../app/middleware/state_enforcer.js'))
+'use strict'
 
-var sinon = require('sinon')
+const path = require('path')
+const assert = require('assert')
+const expect = require('chai').expect
+const stateEnforcer = require(path.join(__dirname, '/../../app/middleware/state_enforcer.js'))
+
+const sinon = require('sinon')
 
 describe('state enforcer', function () {
-  var response = {
+  const response = {
     status: function () {},
     render: function () {}
   }
-  var status
-  var render
-  var next
+  let status
+  let render
+  let next
 
   beforeEach(function () {
     status = sinon.stub(response, 'status')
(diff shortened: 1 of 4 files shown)


test/test_helpers/test_helpers.js
```diff
@@ -106,7 +106,11 @@ function rawSuccessfulGetChargeDebitCardOnly (status, returnUrl, chargeId, gatew
   return charge
 }
 
-function rawSuccessfulGetCharge (status, returnUrl, chargeId, gatewayAccountId, auth3dsData = {}, emailSettings, disableBillingAddress, paymentProvider = 'sandbox') {
+function rawSuccessfulGetCharge (status, returnUrl, chargeId, gatewayAccountId, auth3dsData = {}, emailSettings, disableBillingAddress) {
+  return rawSuccessfulGetChargeWithPaymentProvider(status, returnUrl, chargeId, gatewayAccountId, auth3dsData, 'sandbox', emailSettings, disableBillingAddress)
+}
+
+function rawSuccessfulGetChargeWithPaymentProvider (status, returnUrl, chargeId, gatewayAccountId, auth3dsData = {}, paymentProvider = 'sandbox', emailSettings, disableBillingAddress) {
```
danworth: I'm assuming this is linked to avoiding confusion that led to the recent issue but I'm not sure I see the benefit just yet.
The new `rawSuccessfulGetChargeWithPaymentProvider` does the same as the original `rawSuccessfulGetCharge` albeit with a slightly different ordering of its arguments. Calling the updated `rawSuccessfulGetCharge` is the same as calling the original without providing `paymentProvider` but the defaulting of the `paymentProvider` was to me clearer in the original because it made use of the default arg syntax within the method signature. I guess this is about wanting a more descriptive name or differentiating between explicitly passing a `paymentProvider` and not? The original always returned a charge with a `paymentProvider` with the option to leave it as default or set it, just as the new `rawSuccessfulGetChargeWithPaymentProvider` does.

Maybe I'm not seeing the benefit, I'm happy to chat.

georgeracu: `rawSuccessfulGetChargeWithPaymentProvider` accepts an extra parameter `paymentProvider`. `rawSuccessfulGetCharge` doesn't accept `paymentProvider`, just defaults it to `sandbox`. This method was necessary in tests where the payment provider is Stripe and we need to check in the state enforcer for states for Stripe. We didn't want to change existing functionality so we left the original method as it was. The ordering of the arguments is slightly different as JS doesn't support named parameters but it supports default values. So if we leave last the params that were sometimes `null`, then we will pass only the first ones that are not null, then JS will add the ones with default values and the last ones will be nulled. 
Also we didn't find a way to do method overloading in JS that's why they are two methods with different names.
Another option would have been to add the new param last and give it a default value, but then we would have had to pass the `emailSettings, disableBillingAddress` as `undefined` for each call of this method, due to lack of named parameters.

danworth: OK thanks for the explanation 👍

test/middleware/state_enforcer_test.js
```diff
@@ -1,18 +1,20 @@
...
+const path = require('path')
+const assert = require('assert')
+const expect = require('chai').expect
+const stateEnforcer = require(path.join(__dirname, '/../../app/middleware/state_enforcer.js'))
+
+const sinon = require('sinon')
 
 describe('state enforcer', function () {
-  var response = {
+  const response = {
```
danworth: 👍

PR #614, opened 2019-01-30
PP-4747 - JS error blocking rest of scripts

jonheslop:
This stops it causing errors further down the line.

Noticed becuase the JS wasn’t firing that creates a pseudo pageview we use to count successful transactions

### DIFF
--- a/app/assets/javascripts/browsered/web-payments/helpers.js
+++ b/app/assets/javascripts/browsered/web-payments/helpers.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const allowedCardTypes = window.Card
+const allowedCardTypes = window.Card || {}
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/browsered/web-payments/helpers.js
```diff
@@ -1,7 +1,7 @@
 'use strict'
 
-const allowedCardTypes = window.Card
-const { email_collection_mode } = window.Charge
+const allowedCardTypes = window.Card || {}
+const { email_collection_mode } = window.Charge || {}
```
sfount: This line will still result in `undefined` if window.Charge doesn't exit - I don't think this is meaningful

PR #622, opened 2019-01-31
PP-4641 - Integrate Google Pay with Payment Request API 

jonheslop:
Previously google pay was really just a vanilla payment request API call.

Refactored to use Google Pay as the payment method using - https://developers.google.com/pay/api/web/guides/paymentrequest/tutorial

Extended helper.js to build the Google request and moved out the common bits to keep things DRY

I’ve gone as far as initiating an Google Pay payment and getting auth and a token back from Google but I haven’t done anything with yet

Also tidied up a few things in the markup to make the definition between
google and apple clearer.

### DIFF
--- a/app/assets/javascripts/browsered/web-payments/google-pay.js
+++ b/app/assets/javascripts/browsered/web-payments/google-pay.js
@@ -1,37 +1,55 @@
 'use strict'
 
-const { prepareRequestObject } = require('./helpers')
+const { getGooglePaymentsConfiguration, showErrorSummary } = require('./helpers')
+const { email_collection_mode } = window.Charge
 
-module.exports = () => {
-  const { supportedInstruments, details, options } = prepareRequestObject('standard')
-  const request = new PaymentRequest(supportedInstruments, details, options)
 
-  request.show()
-    .then(result => {
-      const payload = result.toJSON()
-      payload.csrfToken = document.getElementById('csrf').value
-      payload.chargeId = window.paymentDetails.chargeID
-      return fetch(`/payment-request/${window.paymentDetails.chargeID}`, {
-        method: 'POST',
-        redirect: 'follow',
-        credentials: 'same-origin',
-        headers: {
-          'Content-Type': 'application/json'
-        },
-        body: JSON.stringify(payload)
-      }).then(response => {
-        if (response.status === 200) {
-          result.complete('success')
-          return window.location.href = response.url
-        } else {
-          result.complete('fail')
-          return window.location.href = response.url
-        }
-      }).catch(() => {
-        result.complete('fail')
-        return window.location.href = response.url
-      })
-    }).catch(err => {
-      return err
+const processPayment = paymentData => {
+  // show returned data in developer console for debugging
+  console.log(paymentData);
+  // @todo pass payment token to your gateway to process payment
+  const paymentToken = paymentData.details.paymentMethodData.tokenizationData.token;
+  console.log('paymentToken', paymentToken);
+}
+
+const createGooglePaymentRequest = () => {
+  const methodData = [{
+    supportedMethods: 'https://google.com/pay',
+    data: getGooglePaymentsConfiguration()
+  }];
+
+  const details = {
+    total: {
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/browsered/web-payments/google-pay.js
```diff
@@ -1,37 +1,55 @@
...
+      }
+    }
+  };
+
+  const options = {
+    requestPayerEmail: email_collection_mode !== 'OFF',
+    requestPayerName: true
+  };
+
+  return new PaymentRequest(methodData, details, options);
```
oswaldquek: does this PaymentRequest make the call to to google pay API?

app/views/includes/scripts.njk
```diff
@@ -51,6 +51,10 @@
       analyticsTrackConfirmClick().init('{{analytics.analyticsId}}', '{{analytics.type}}', '{{analytics.paymentProvider}}', '{{analytics.amount}}', '{{hitPage}}');
     }
   });
+  {% if allowWebPayments %}
+    window.googlePayGatewayMerchantID = '{{ googlePayGatewayMerchantID }}';
+    window.googlePayMerchantID = '{{ googlePayMerchantID }}';
```
oswaldquek: i guess we always get these google pay specific things even if the user only has apple pay enabled on the device?

app/assets/javascripts/browsered/web-payments/google-pay.js
jonheslop: Yes this launches the native dialog and because here — 

https://github.com/alphagov/pay-frontend/pull/622/files#diff-2b4385fbba492188502b49e1137f1efbR17 

Google Pay is declared as a payment method, Chrome knows how to contact Google Pay

app/views/includes/scripts.njk
jonheslop: Yeah that’s fine we dump loadsa bits down here just so that the client side script can grab hold of them if it needs them.

app/assets/javascripts/browsered/web-payments/index.js
```diff
@@ -1,36 +1,46 @@
...
+const { clearErrorSummary } = require('./helpers')
 const makeApplePayRequest = require('./apple-pay')
-const makePaymentRequest = require('./google-pay')
+const { createGooglePaymentRequest, googlePayNow } = require('./google-pay')
 
 module.exports = () => {
   const paymentMethodForm = document.getElementById('payment-request-container')
   const standardMethodContainer = document.getElementById('enter-card-details-container')
 
   if (window.PaymentRequest && paymentMethodForm) {
```
oswaldquek: what does `paymentMethodForm` do?

app/assets/javascripts/browsered/web-payments/index.js
```diff
@@ -1,36 +1,46 @@
...
+      const request = createGooglePaymentRequest();
+      request.canMakePayment()
+        .then(result => {
+          if (result) {
+            console.log('Google Pay is available')
+            paymentMethodForm.classList.remove('hidden')
+            standardMethodContainer.classList.add('hidden')
+            document.getElementById('payment-method-google-pay').parentNode.style.display = 'block'
+          }
+        })
```
oswaldquek: i take it all the above is saying the user can pay with either Apple Pay or G Pay but not both?

app/assets/javascripts/browsered/web-payments/index.js
jonheslop: its an element on the page, it’s only there if the service has made webpayments available (a boolean stored in the DB) so this is how the clientside JS works out whether to trigger for webpayments, it’s stored as a `const` because it’s quicker to use the variable than look for the element everytime we want to do something with it.

app/assets/javascripts/browsered/web-payments/index.js
jonheslop: Yeah so we know webpayments are available on the service but then we check the browser is enabled to do them, then we trigger Apple Pay if it’s Safari or else Google Pay

PR #630, opened 2019-02-07
PP-4784 init Google / Apple Pay only if configured

jonheslop:
a7eb620 - PP-4784 - Put scripts into window so can conditionally trigger
4febb7c - PP-4784 - introduce concept of enabling Apple / Google Pay individually
* Using allowApplePay and allowGooglePay so we can toggle individually. This hasn’t been done in connector so fudging with change in charge_controller.js
* Split up the initalisation of web payments into different functions and toggling only if browser has the capability.

### DIFF
--- a/app/assets/javascripts/browsered/web-payments/index.js
+++ b/app/assets/javascripts/browsered/web-payments/index.js
@@ -4,7 +4,7 @@ const { clearErrorSummary } = require('./helpers')
 const makeApplePayRequest = require('./apple-pay')
 const { createGooglePaymentRequest, googlePayNow } = require('./google-pay')
 
-module.exports = () => {
+const init = () => {
   const paymentMethodForm = document.getElementById('payment-request-container')
   const standardMethodContainer = document.getElementById('enter-card-details-container')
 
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/browsered/web-payments/index.js
```diff
@@ -1,33 +1,41 @@
...
+    paymentMethodForm.classList.remove('hidden')
+    standardMethodContainer.classList.add('hidden')
+    document.getElementById('payment-method-apple-pay').parentNode.style.display = 'block'
+    document.getElementById('payment-method-apple-pay').checked = true
+  }
+}
+
+const initGooglePayIfAvailable = () => {
+  if (window.PaymentRequest) {
+    console.log('payment request API is available')
```
gidsg: I guess we shouldn't leave these `console.log` instances in production, could we at least add a `TODO` to remove them before we go live?

jonheslop: Yeah I left them all in because it’s all behind the feature flag and there is still a few bits to do to prepare this for production around getting vars and stuff, I guess wouldn’t do any harm if I added TODOs to help make it clear I will remove

PR #653, opened 2019-02-18
PP-4754 Cypress card flow

sfount:
E2E tests for standard card payment flow. 
* build infrastructure and utilities for stubbing connector, admin users, and card id responses 
* spec file for payments 
* cookie and session management


jonheslop:
Just need to address the Standard lint errors

PR #655, opened 2019-02-18
BAU - Android complaining about missing object that docs said we shou…

jonheslop:
…ldnt include

### DIFF
--- a/app/assets/javascripts/browsered/web-payments/helpers.js
+++ b/app/assets/javascripts/browsered/web-payments/helpers.js
@@ -110,12 +110,18 @@ const getGooglePaymentsConfiguration = () => {
   }
 
   return {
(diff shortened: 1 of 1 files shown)


mrlumbu:
can we not BAU this - it is a change to how we process live payments

oswaldquek:
@jonheslop can you include these in the commit message:

- the link to https://developers.google.com/pay/api/web/guides/tutorial#paymentdatarequest which explains why you've added it
- the error message we got when debugging: `WalletMerchantError: Error in loadWebPaymentData: PaymentDataRequest.transactionInfo is required!`
- the story number which is PP-4644 as we're doing this to get G Pay into prod

jonheslop:
Will open another with the Jira ticket

PR #668, opened 2019-02-22
PP-4806 Check pact compatibility and tag pact during build

kbottla:
## WHAT
- Check pact compatibility and tag pact (test) during master build

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -111,12 +111,22 @@ pipeline {
         branch 'master'
       }
       steps {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -111,13 +111,23 @@ pipeline {
         branch 'master'
       }
       steps {
+        checkPactCompatibility("frontend", gitCommit(), "test")
```
danworth: Looks sensible to me but probably a good idea for Oz to review and agree too.

PR #669, opened 2019-02-22
PP-4890: Add logo for Somerset West and Taunton Council

rhowe-gds:
## WHAT
Add an SVG version of Taunton Deane Borough Council & West Somerset Council's logo


jonheslop:
We also need to add a custom stylesheet to go with this SVG. Will contact them to see what this should look like.

PR #671, opened 2019-02-25
PP-4816 - Apple Pay Cypress tests

jonheslop:
Fairly similar to the Google Pay one, but `ApplePaySesssion` is a class so had to construct mock class to replicate behaviour client-side JS expects


simad:
Best tests ever 💯 🥇

PR #687, opened 2019-02-28
PP-4927 - maintain web payments flow after auth waiting

jonheslop:
Sometimes a card cannot be auth’d immediately, in this case 202/409 is return and [we wait](https://github.com/alphagov/pay-frontend/blob/master/app/views/auth_waiting.njk#L27).

Once the auth is complete the flow went back to non Apple/Google Pay flow which meant users would see the confirm page, which we dont want.

This adds logic to check if it’s a digital wallet payment and if so immediately does the auth and then redirects to the service’s returnURL

### DIFF
--- a/app/controllers/charge_controller.js
+++ b/app/controllers/charge_controller.js
@@ -212,7 +212,22 @@ module.exports = {
         redirect(res).toAuth3dsRequired(req.chargeId)
         break
       default:
-        redirect(res).toConfirm(req.chargeId)
+        if (charge.walletType !== undefined) {
+          Charge(req.headers[CORRELATION_HEADER])
+            .capture(req.chargeId)
+            .then(
+              () => redirect(res).toReturn(req.chargeId),
+              err => {
+                if (err.message === 'CAPTURE_FAILED') return responseRouter.response(req, res, 'CAPTURE_FAILURE', withAnalytics(charge))
+                responseRouter.response(req, res, 'SYSTEM_ERROR', withAnalytics(
+                  charge,
+                  { returnUrl: routeFor('return', charge.id) }
+                ))
+              }
+            )
+        } else {
+          redirect(res).toConfirm(req.chargeId)
+        }
     }
   },
   confirm: (req, res) => {
--- a/test/integration/charge_ft_tests.js
+++ b/test/integration/charge_ft_tests.js
@@ -396,6 +396,25 @@ describe('chargeTests', function () {
           .end(done)
       })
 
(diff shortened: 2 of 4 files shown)


app/controllers/charge_controller.js
```diff
@@ -212,7 +212,22 @@ module.exports = {
         redirect(res).toAuth3dsRequired(req.chargeId)
         break
       default:
-        redirect(res).toConfirm(req.chargeId)
+        if (charge.walletType !== undefined) {
```
danworth: My JS is a bit rusty but if `charge.walletType` is `null` it'll pass this check won't it?
```
> null !== undefined
true
```
 I ask because Connector should always return `wallet_type`, it'll be `null` for non-wallet payments else `GOOGLE_PAY || APPLE_PAY`.  I think we need to check that it isn't a) there but set to null b) not there at all (just to be safe). I guess it depends how Connector response is deserialised and if `wallet_type: null` and `wallet_type` not existing at all becomes `walletType = undefined` on the charge object 🤷‍♂️ .

danworth: Actually our current `ChargeResponse` does not include null values, those keys will be omitted. Perhaps we should follow this pattern (recommended by google style guide). In that case `wallet_type` will only be present if its a wallet payment with values `APPLE_PAY | GOOGLE_PAY`. Probably still need to consider the scenario where the key exists and its value is `null` since if our design changes to include `null` values we don't want to accidentally skip confirm page for card payments.

jonheslop: I think this is covered by 
https://github.com/alphagov/pay-frontend/blob/28c643cd5cb0007adc280ba2a0f7c3d69a90fa1c/app/services/normalise_charge.js#L35-L37

jonheslop: The normailiser will omit it, so if it’s `null` it will always be `undefined` as the controller sees it

danworth: ah cool 👍

test/integration/charge_ft_tests.js
```diff
@@ -396,6 +396,25 @@ describe('chargeTests', function () {
           .end(done)
       })
 
+      it('should redirect user from /auth_waiting to the returnURL when connector returns a successful status and is web payment (Apple/Google Pay)', function (done) {
```
danworth: Nice additional test, plus I see there's already one to confirm it redirects to confirm page for non wallet payments 👍

PR #691, opened 2019-03-04
[WIP] PP-4896: Apple Pay domain validation using an environment variable rather than hardcoded

46bit:
## What

Adds a `APPLE_PAY_MERCHANT_ID_DOMAIN_ASSOCIATION` environment variable. In every Pay environment this will be set to the latest text that Apple want us to serve from `/.well-known/apple-developer-merchantid-domain-association.txt`. We believe that Apple periodically request that URL (every 6-ish months) and so hardcoding one environment's value (as done until now) has too much to toil.

## How to review

* Code review;
* See if the tests go green;
* 🐝 Before merging wait for :
  * [ ] A `pay-low-pass` PR containing this secret;
  * [ ] A `pay-infra` PR that provides the secret to frontend.


test/controllers/web-payments/apple-pay/merchantid-domain-association-controller-test.js
```diff
@@ -0,0 +1,40 @@
+var path = require('path')
```
jonheslop: Every `var` and `let` here could be replace by a `const`

46bit:
⚠️ This won't work yet. `tf secrets provision` requires secrets to be on one line when uploading to AWS Parameter Store, but Apple's file is on multiple lines. I'm planning to overhaul our secrets management to use Terraform at which point I think we will support multi-line secrets, but that's a week or few away.

I'll leave this PR open so we can merge it once that work is done.

app/controllers/web-payments/apple-pay/merchantid-domain-association-controller.js
```diff
@@ -0,0 +1,16 @@
+'use strict'
+
+// NPM dependencies
+const logger = require('winston')
+
+// Local constants
+const { APPLE_PAY_MERCHANT_ID_DOMAIN_ASSOCIATION } = process.env
+
+module.exports = (req, res) => {
+  if (APPLE_PAY_MERCHANT_ID_DOMAIN_ASSOCIATION.length === 0) {
```
sfount: If this environment variable doesn't exist for some reason this will throw an error that isn't handled (likely resulting in 500 stack trace sent to the client). 

Options: 
* Short circuit evaluation to make sure it exists as part of the evaluation
* Check to make sure environment is set up at the top and use that logic in the controller

Short circuit option: 
```suggestion
  if (APPLE_PAY_MERCHANT_ID_DOMAIN_ASSOCIATION && APPLE_PAY_MERCHANT_ID_DOMAIN_ASSOCIATION.length === 0) {
```

46bit: > If this environment variable doesn't exist for some reason this will throw an error that isn't handled (likely resulting in 500 stack trace sent to the client).

I tested this and was surprised: I found there was no error and the `APPLE_PAY_MERCHANT_ID_DOMAIN_ASSOCIATION` constant was empty.

46bit: Approved in any case–it does look less dangerous 💯

46bit:
Closing until/if we find time to change how Pay manages secrets.

PR #708, opened 2019-03-07
PP-4917 - Adding GA events for Apple Pay errors

jonheslop:
- If auth fails with connector
- If auth fails to push to frontend controller


46bit:
The logic for how this decides whether Frontend or Connector was the problem is quite implicit and I can't say for sure how it does it.

jonheslop:
Good point updated messaging

PR #710, opened 2019-03-07
PP-4904 lint browser javascripts too

jonheslop:
Fixed all warnings

Would recommend viewing without whitespace changes — https://github.com/alphagov/pay-frontend/pull/710/files?w=1

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,10 +18,15 @@
       "it",
       "expect",
       "cy",
-      "Cypress"
+      "Cypress",
+      "i18n",
+      "PaymentRequest",
+      "ApplePaySession",
+      "ga",
+      "fetch"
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -18,10 +18,15 @@
       "it",
       "expect",
       "cy",
-      "Cypress"
+      "Cypress",
+      "i18n",
+      "PaymentRequest",
+      "ApplePaySession",
```
sfount: This looks like it adds 5 dependencies, are these supposed to be in this PR?

sfount: EDIT - didn't look at context 🤦‍♂️

jonheslop: 😅

PR #713, opened 2019-03-08
PP-4939 Fix error when service has organisation name but not address

stephencdaly:
- It is possible to set individual elements of the merchant details
- Previously service model was assuming that if the merchant_details field was present in the get service response, it had a full address and so attempted to resolve the country from the country code without checking whether it was present. This commit fixes this.
- Handle the case where there is an organisation name but no address or an incomplete address by displaying just the organisation name in the footer of the payment pages.

Co-authored-by: Yancoba Thompson <yancoba.thompson@digital.cabinet-office.gov.uk>

### DIFF
--- a/app/models/Service.class.js
+++ b/app/models/Service.class.js
@@ -1,8 +1,5 @@
 'use strict'
 
-// NPM dependencies
-const lodash = require('lodash')
-
 // Local dependencies
 const countries = require('../services/countries')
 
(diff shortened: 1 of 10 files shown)


app/models/Service.class.js
```diff
@@ -38,29 +38,15 @@ class Service {
       addressLine2: serviceData.merchant_details.address_line2,
       city: serviceData.merchant_details.address_city,
       postcode: serviceData.merchant_details.address_postcode,
-      countryName: countries.translateCountryISOtoName(serviceData.merchant_details.address_country)
+      countryName: serviceData.merchant_details.address_country ? countries.translateCountryISOtoName(serviceData.merchant_details.address_country) : undefined
     } : undefined
-  }
 
-  /**
```
stephencdaly: removed the below methods as they were unused

app/models/Service.class.js
```diff
@@ -38,29 +35,15 @@ class Service {
...
-   *
-   * All services must have merchant details defined for all gateway accounts
-   * This will be kept as backward compatibility until merchant details are
-   * enforced.
-   *
-   * @returns {boolean} if the service got merchant details specified
-   */
-  hasMerchantDetails () {
-    return !lodash.isEmpty(this.merchantDetails)
+    this.hasCompleteMerchantDetailsAddress = !!(
```
DanailMinchev: 👍

PR #717, opened 2019-03-11
PP-4890 Fix Taunton Deane custom branding logo

stephencdaly:
Logo had an invalid format, add in xlink definition to make it valid

I added in  ```xmlns:xlink="http://www.w3.org/1999/xlink"``` to the svg tag. This fixes it, but should the xlink elements be present in the first place?


jonheslop:
🙈 Aww no Stephen, sorry I didnt see this!?

PR #718, opened 2019-03-12
BAU - remove masks and xlinks

jonheslop:
Not sure what these things in the SVG were but they somehow corrupted it when I ran the SVG through [SVGOMG](https://jakearchibald.github.io/svgomg/)


sfount:
Hard to see exactly what happened here but it's a small change and should be verifiable manually.

jonheslop:
It had these in there
```xml
 <mask id="mask-4" fill="white">
   <use xlink:href="#path-3"></use>
  </mask>
```
And they caused this
Screen Shot 2019-03-12 at 09 51 58
I took them out and now it works fine
Screen Shot 2019-03-12 at 09 52 31

jonheslop:
probably not will check and remove if so

jonheslop:
For some reason logo reminds me of this

Liverpool FC Away kit 1992 - 1993

PR #719, opened 2019-03-12
PP-4916 success monitoring for google pay

jonheslop:
Logs on client and server side when Google Payment goes well, as server side we use same controllers for Apple and Google, these logs will partly resolve PP-4917 too

### DIFF
--- a/app/assets/javascripts/browsered/web-payments/google-pay.js
+++ b/app/assets/javascripts/browsered/web-payments/google-pay.js
@@ -13,6 +13,7 @@ const processPayment = paymentData => {
     body: JSON.stringify(paymentData)
   })
     .then(response => {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/browsered/web-payments/google-pay.js
```diff
@@ -13,6 +13,7 @@ const processPayment = paymentData => {
     body: JSON.stringify(paymentData)
   })
     .then(response => {
+      ga('send', 'event', 'Google Pay', 'Sucessful', 'auth/capture request')
```
oswaldquek: i take it this section of code handles authorisation and capture, and not just authorisation?

jonheslop: Yeah it waits for the return from both bits

PR #721, opened 2019-03-12
BAU - use minified JS

jonheslop:
This should have never have been changed 🙈
I think Silvia and I were just making it easier to debug but shouldn’t
have ever shipped to production 🤦‍♂️


jonheslop:
TBF I think it was almost certainly me

jonheslop:
But happy to scapegoat @simad as she left us

jonheslop:
Yeah nice idea

simad:
@alexbishop1 
bye

PR #722, opened 2019-03-12
PP-4924 - CSS is minified so we should name it so

jonheslop:
All our other repos follow this convention of naming the CSS min.css

But to make sure whilst mid-deploy if a server requests either we will
generate both


alexbishop1:
Regarding “But to make sure whilst mid-deploy if a server requests either we will generate both”: isn’t there still the risk that one frontend node returns a reference to the `.min.css` file but the request to retrieve it will hit a node that does not yet have the `.min.css` file?

alexbishop1:
I guess I’m saying that maybe we want to have one release that generates the file in both locations and then another release that makes us use the new location.

jonheslop:
Ah yeah I thought of backwards compat but not forwards ok will remove server.js change

PR #729, opened 2019-03-13
PP-4924 - remove legacy assets

jonheslop:
Pay JS commons now handles custom branding so we don’t need to keep the assets in fronted, but instead can fetch them from JS commons


jonheslop:
Yeah it does. I see this as a middle solution, we want custom branding for payment links and this fulfils that, but yeah if custom branding gets bigger this is gonna be annoying, so I think something like I’ve mentioned before about it putting assets on S3 and adding links to them in the DB might work better.

PR #732, opened 2019-03-14
PP-4917 - remove logging line

jonheslop:


jonheslop:
Perhaps we hold off merging until we’re 100% cofident we’ve sorted the NAXSI issues

PR #733, opened 2019-03-14
PP-4917 - Make card brand normalisation independent of case

jonheslop:
Apple have docs about how to declare which card types you want to accept, but not how it labels those card types coming out, basically it capitalises them, so adding `toLowerCase()` to account for this.

### DIFF
--- a/app/controllers/web-payments/apple-pay/normalise-apple-pay-payload.js
+++ b/app/controllers/web-payments/apple-pay/normalise-apple-pay-payload.js
@@ -4,8 +4,9 @@
 const humps = require('humps')
 
 const normaliseCardName = cardName => {
-  switch (cardName) {
-    case 'MasterCard':
+  const lowerCaseCardName = cardName.toLowerCase()
+  switch (lowerCaseCardName) {
+    case 'mastercard':
       return 'master-card'
     case 'amex':
       return 'american-express'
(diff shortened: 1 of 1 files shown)


app/controllers/web-payments/apple-pay/normalise-apple-pay-payload.js
```diff
@@ -15,19 +15,21 @@ const normaliseCardName = cardName => {
...
 }
+
 const normaliseLastDigitsCardNumber = displayName => displayName.substr(displayName.length - 4)
 
 module.exports = payload => {
   const paymentInfo = {
     last_digits_card_number: normaliseLastDigitsCardNumber(payload.token.paymentMethod.displayName),
-    brand: normaliseCardName(payload.token.paymentMethod.network),
-    card_type: payload.token.paymentMethod.type.toLocaleUpperCase(),
+    brand: normaliseCardName(payload.token.paymentMethod.network.toLowerCase()),
```
sfount: Would it be a good idea to move the `toLowerCase()` call into `normaliseCardName`? That way it wouldn't be dependent on the code that is calling it to know it has to make it lower case

jonheslop: Yeah good point

PR #737, opened 2019-03-14
PP-4917 performance improvements for apple pay

jonheslop:
Moving the check to see if Apple Pay is available to inline, this way it happens quicker so there is a less of a flash of things jumping around

### DIFF
--- a/app/assets/sass/modules/_payment-request.scss
+++ b/app/assets/sass/modules/_payment-request.scss
@@ -1,7 +1,3 @@
-.payment-request-radios .govuk-radios__item:not(:last-child) {
-  display: none;
-}
-
 .payment-request-logo {
   float: left;
   margin-top: -.25rem;
@@ -9,3 +5,42 @@
   max-height: 2rem;
   width: auto;
 }
+
+.web-payments-container {
+  display: none;
+}
+
+// Toggle Apple Pay
+.apple-pay-available {
+  #web-payments-container {
+    display: block;
+  }
+
+  #enter-card-details-container {
+    display: none;
+  }
+}
+
+.apple-pay-unavailable {
+  .payment-request-radios .govuk-radios__item:first-child {
+    display: none;
+  }
+}
+
+// Toggle Google Pay
+.google-pay-available {
+  #web-payments-container {
+    display: block;
+  }
+
+  #enter-card-details-container {
+    display: none;
+  }
+}
+
+.google-pay-unavailable {
+  .payment-request-radios .govuk-radios__item:nth-child(2) {
+    display: none;
+  }
+}
+
--- a/app/views/charge.njk
+++ b/app/views/charge.njk
@@ -78,6 +78,7 @@
           }
         })
       }}
+      <div id="spinner" class="hidden"><img src="/images/spinner.gif" alt="{{ __("authorisation.spinnerAltText") }}"/></div>
     </form>
   {% endif %}
 
--- a/app/views/charge.njk
+++ b/app/views/charge.njk
@@ -6,7 +6,20 @@
   {{ __("cardDetails.title") }}
 {% endblock %}
 
+{% set bodyClasses = "google-pay-unavailable" %}
 {% set mainClasses = "charge-new" %}
+
+{% if allowApplePay%}
+  {% block bodyStart %}
+    <script>
+      if (window.ApplePaySession && window.ApplePaySession.canMakePayments()) {
+        document.body.classList.add('apple-pay-available')
+      } else {
+        document.body.classList.add('apple-pay-unavailable')
+      }
+    </script>
+  {% endblock %}
+{% endif %}
 
 {% block content %}
 <div class="govuk-grid-row govuk-!-margin-bottom-9">
@@ -27,7 +40,7 @@
     </div>
 
   {% if allowApplePay or allowGooglePay %}
-    <form id="payment-request-container" class="hidden">
+    <form id="web-payments-container">
       {{ govukRadios({
         name: "payment-method",
         classes: "payment-request-radios",
--- a/app/views/charge.njk
+++ b/app/views/charge.njk
@@ -17,6 +17,13 @@
       } else {
         document.body.classList.add('apple-pay-unavailable')
       }
+
+      {# If our JS bundle fails to load we need to undo the above Apple Pay check #}
+      document.addEventListener('DOMContentLoaded', function(event) {
+        if (!window.payScripts.webPayments) {
+          document.body.classList.remove('apple-pay-available')
+        }
+      })
     </script>
   {% endblock %}
 {% endif %}
--- a/app/views/includes/scripts.njk
+++ b/app/views/includes/scripts.njk
@@ -58,19 +58,3 @@
   });
 </script>
 <script src="{{ js_path }}"></script>
-<script>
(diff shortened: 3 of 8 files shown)


app/assets/sass/modules/_payment-request.scss
```diff
@@ -1,11 +1,23 @@
...
 .payment-request-logo {
   float: left;
   margin-top: -.25rem;
   margin-right: govuk-spacing(3);
   max-height: 2rem;
   width: auto;
 }
+
+.apple-pay-available {
+  #payment-request-container.hidden {
```
jonheslop: This is gross, can improve

app/views/charge.njk
```diff
@@ -6,8 +6,13 @@
   {{ __("cardDetails.title") }}
 {% endblock %}
 
+{% set bodyClasses = "google-pay-unavailable" %}
```
alexbishop1: Do we actually need both these `-available` and `unavailable` classes? Can’t we just have the latter and use its absence to denote unavailability?

app/views/charge.njk
```diff
@@ -6,8 +6,13 @@
...
 {% set mainClasses = "charge-new" %}
 
+{% block bodyStart %}
```
alexbishop1: If there `apple-pay-unavailable` weren’t a thing, this code could be wrapped in an `{ %if allowApplePay %}` block

app/views/charge.njk
```diff
@@ -6,8 +6,13 @@
...
+  <script>document.body.className = document.body.className + ((window.ApplePaySession && window.ApplePaySession.canMakePayments()) ? ' apple-pay-available' : ' apple-pay-unavailable');</script>
```
alexbishop1: Any reason not to use `classList`? We’re already relying on it for reading classes from elements

app/views/charge.njk
```diff
@@ -27,7 +32,7 @@
     </div>
 
   {% if allowApplePay or allowGooglePay %}
-    <form id="payment-request-container" class="hidden">
+    <form class="web-payments-container">
```
alexbishop1: Why did this change from being an ID to a class?

jonheslop: ID’s are good for JS to hook into, but bad (and slower) to use in CSS, so I switched

app/views/charge.njk
jonheslop: Yeah good point

app/views/charge.njk
jonheslop: Allowing Apple Pay and Apple Pay being unavailable are different things

app/views/charge.njk
jonheslop: It makes the CSS easier and less overiding this way

app/views/charge.njk
jonheslop: Updated PR

app/views/charge.njk
alexbishop1: Yes but what I meant was that we don’t need to check if Apple Pay is available if it’s not allowed in the first place

app/views/charge.njk
alexbishop1: I thought it was the other way round…

https://csswizardry.com/2011/09/writing-efficient-css-selectors/

app/views/includes/scripts.njk
```diff
@@ -57,3 +57,19 @@
   });
 </script>
 <script src="{{ js_path }}"></script>
+<script>
+{# If our JS bundle fails to load we need to undo Apple Pay check #}
+function ready(fn) {
+  if (document.attachEvent ? document.readyState === "complete" : document.readyState !== "loading"){
+    fn();
+  } else {
+    document.addEventListener('DOMContentLoaded', fn);
```
alexbishop1: Not sure we should be mixing `DOMContentLoaded` and `document.readyState`. If my recollection of https://html.spec.whatwg.org/multipage/parsing.html#the-end is correct, `DOMContentLoaded` fires before `document.readyState` is set to `"completed"`.

Could we have a check to see if `document.readyState` is "interactive" or "complete" and, if it isn’t, set an listener to listen for the `readystatechange` change event and do the check again?

app/views/includes/scripts.njk
```diff
@@ -57,3 +57,19 @@
...
+  }
+}
+
+ready(() => {
```
alexbishop1: Are we OK to use arrow functions? Presumably JavaScript inside Nunjucks templates doesn’t go through the compiler as JavaScript in external files would.

app/views/includes/scripts.njk
jonheslop: This makes it work in more browsers — http://youmightnotneedjquery.com/#ready

app/views/includes/scripts.njk
jonheslop: OOF! Good point

app/views/includes/scripts.njk
```diff
@@ -57,3 +57,19 @@
   });
 </script>
 <script src="{{ js_path }}"></script>
+<script>
+{# If our JS bundle fails to load we need to undo Apple Pay check #}
+function ready(fn) {
+  if (document.attachEvent ? document.readyState === "complete" : document.readyState !== "loading"){
```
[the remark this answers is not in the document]
jonheslop: I just copied it from here http://youmightnotneedjquery.com/#ready

app/views/includes/scripts.njk
alexbishop1: Hmm, still not sure we should be mixing `DOMContentLoaded` and `readyState` like this. I would have thought that browsers that support `document.readyState` would also support the `readystatechange` event.

Would it be possible to not have the second inline `script` element at all and instead have the first inline `script` element do its thing then set an event listener to listen for the `DOMContentLoaded` event, defining the function that should run right there? That would eliminate the need to check the `readyState` at all because it would also run when `DOMContentLoaded` fires.

app/views/includes/scripts.njk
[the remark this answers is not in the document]
jonheslop: I’ve updated it all now, could you rereview?

app/views/includes/scripts.njk
jonheslop: I’ve updated it all now, could you rereview?

app/views/charge.njk
```diff
@@ -6,8 +6,28 @@
...
+        document.body.classList.add('apple-pay-unavailable')
+      }
+
+      {# If our JS bundle fails to load we need to undo the above Apple Pay check #}
+      document.addEventListener('DOMContentLoaded', function(event) {
+        if (!window.payScripts.webPayments) {
+          document.body.classList.remove('apple-pay-available')
+        }
+      })
+    </script>
```
alexbishop1: This is much nicer

app/views/charge.njk
```diff
@@ -27,7 +47,7 @@
     </div>
 
   {% if allowApplePay or allowGooglePay %}
-    <form id="payment-request-container" class="hidden">
+    <form id="web-payments-container">
```
alexbishop1: Changing the ID in the HTML and the CSS in the external stylesheet is risky (because one can end up with the new HTML but the old CSS or vice-versa) but since no live services have Apple Pay or Google Pay enabled yet, I think this should be OK

PR #747, opened 2019-03-19
PP-4653 - Check if service is live to set Google Pay environment

jonheslop:


app/assets/javascripts/browsered/web-payments/helpers.js
```diff
@@ -108,7 +108,7 @@ const getGooglePaymentsConfiguration = () => {
   }
 
   return {
-    environment: window.googlePayMerchantID.length > 0 ? 'PRODUCTION' : 'TEST',
+    environment: (window.googlePayMerchantID.length > 0 && window.gatewayAccountType !== 'test') ? 'PRODUCTION' : 'TEST',
```
oswaldquek: would it be better to make it `window.gatewayAccountType == 'live'`? in the unlikely event of a new gateway account type...

jonheslop: Ah yeah good idea

oswaldquek:
@jonheslop Thanks for the PR. Be sure to test a live payment in prod when it's deployed!

PR #751, opened 2019-03-21
PP-4961: Pact test for Apple Pay

oswaldquek:
Test for authorisation success.

### DIFF
--- a/test/unit/clients/connector_client_apple_authentication_tests.js
+++ b/test/unit/clients/connector_client_apple_authentication_tests.js
@@ -71,4 +71,36 @@ describe('connectors client - apple authentication API', function () {
       })
     })
   })
+
+  describe('authorisation declined', function () {
+    const appleAuthRequest = fixtures.appleAuthRequestDetails({ lastDigitsCardNumber: '0002' })
+    const authorisationDeclinedResponse = fixtures.webPaymentFailedResponse('This transaction was declined.')
+
+    before((done) => {
+      provider.addInteraction(
+        new PactInteractionBuilder(APPLE_AUTH_PATH)
+          .withRequestBody(appleAuthRequest.getPactified())
+          .withMethod('POST')
+          .withState('a sandbox account exists with a charge with id testChargeId that is in state ENTERING_CARD_DETAILS.')
+          .withUponReceiving('a valid apple pay auth request which should be declined')
+          .withResponseBody(authorisationDeclinedResponse.getPactified())
+          .withStatusCode(400)
+          .build()
+      )
+        .catch(done())
+    })
+
+    afterEach(() => provider.verify())
+
+    it('should return authorisation declined', function (done) {
+      connectorClient({ baseUrl: BASEURL }).chargeAuthWithWallet({
+        chargeId: TEST_CHARGE_ID,
+        provider: 'apple',
+        payload: appleAuthRequest.getPlain()
+      }).then(res => {
+        expect(res.body.message).to.be.equal('This transaction was declined.')
+        done()
+      }).catch((err) => done('should not be hit: ' + JSON.stringify(err)))
+    })
+  })
 })
--- a/test/unit/clients/connector_client_apple_authentication_tests.js
+++ b/test/unit/clients/connector_client_apple_authentication_tests.js
@@ -65,7 +65,7 @@ describe('connectors client - apple authentication API', function () {
         }).then(res => {
           expect(res.body.status).to.be.equal('AUTHORISATION SUCCESS')
           done()
-        }).catch((err) => done('should not be hit: ' + JSON.stringify(err)))
+        }).catch((err) => done(new Error('should not be hit: ' + JSON.stringify(err))))
       })
     })
   })
@@ -96,7 +96,7 @@ describe('connectors client - apple authentication API', function () {
       }).then(res => {
         expect(res.body.message).to.be.equal('This transaction was declined.')
         done()
-      }).catch((err) => done('should not be hit: ' + JSON.stringify(err)))
+      }).catch((err) => done(new Error('should not be hit: ' + JSON.stringify(err))))
     })
   })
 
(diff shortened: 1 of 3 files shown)


test/unit/clients/connector_client_apple_authentication_tests.js
```diff
@@ -0,0 +1,138 @@
...
+          new PactInteractionBuilder(APPLE_AUTH_PATH)
+            .withRequestBody(appleAuthRequest.getPactified())
+            .withMethod('POST')
+            .withState('a sandbox account exists with a charge with id testChargeId that is in state ENTERING_CARD_DETAILS.')
+            .withUponReceiving('a valid apple pay auth request which should be authorised')
+            .withResponseBody(authorisationSuccessResponse.getPactified())
+            .withStatusCode(200)
+            .build()
+        )
+          .catch(done())
```
stephencdaly: should be a callback that's passed to catch i.e.
```.catch(done)```
which Steven informed me is equivalent to 
```.catch(err => done(err)```

or
```.catch(() => done())```

File: test/unit/clients/connector_client_apple_authentication_tests.js [hunk 2]
```diff
@@ -0,0 +1,138 @@
...
+    provider: 'connector',
+    port: PORT,
+    log: path.resolve(process.cwd(), 'logs', 'mockserver-integration.log'),
+    dir: path.resolve(process.cwd(), 'pacts'),
+    spec: 2,
+    pactfileWriteMode: 'merge'
+  })
+
+  before(() => provider.setup())
+  after((done) => provider.finalize().then(done()))
```
sfount: Re: @stephencdaly comment https://github.com/alphagov/pay-frontend/pull/751/files#r268110358. Same here!

oswaldquek:
@sfount @stephencdaly I've reverted to `.catch(done())` as `.catch(() => done())` results in the following errors:
```11:52:33   2) connectors client - google authentication API
11:52:33        "after all" hook in "connectors client - google authentication API":
11:52:33      Error: Resolution method is overspecified. Specify a callback *or* return a Promise; not both.``` or
```11:39:02   1) connectors client - apple authentication API
11:39:02        Authenticate apple payment
11:39:02          authorisation success
11:39:02            "before all" hook for "should return authorisation success":
11:39:02      Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/app/test/unit/clients/connector_client_apple_authentication_tests.js)```

**Comment by sfount** [comment]
@oswaldquek This is likely in 
```js
after((done) => provider.finalize().then(done()))
```

Basically if you are returning a `Promise` anyway you do not need to accept the `done` parameter into the callback, for example: 
```js
after(() => provider.finalize())
```

The alternative relying on done would be: 
```js
after((done) => { 
  provider.finalize()
  .then(done)
})
```

I think for cases like this just returning the Promise chain seems cleaner.

stephencdaly:
I think you need
```
.then(done)
.catch(done)
```
as suggested by the error
```Error: Resolution method is overspecified. Specify a callback *or* return a Promise; not both```

Or maybe you just need neither.
I was just looking at this which might clear it up https://mochajs.org/#working-with-promises

I'm also writing some asynchronous code and I'm opting to use the async/await syntax as I think it makes more sense to me than the done callbacks.
https://mochajs.org/#using-async-await

**Comment by stephencdaly** [comment]
I was just reading this, which made things a little clearer around this for me
https://mochajs.org/#asynchronous-code

File: test/unit/clients/connector_client_apple_authentication_tests.js [hunk 3]
```diff
@@ -0,0 +1,132 @@
...
+      it('should return authorisation success', function (done) {
+        const payload = appleAuthRequest.getPlain()
+        connectorClient({ baseUrl: BASEURL }).chargeAuthWithWallet({
+          chargeId: TEST_CHARGE_ID,
+          provider: 'apple',
+          payload: payload
+        }).then(res => {
+          expect(res.body.status).to.be.equal('AUTHORISATION SUCCESS')
+          done()
+        }).catch((err) => done('should not be hit: ' + JSON.stringify(err)))
```
sfount: We might run into `Error: done() invoked with non-Error:` this again if we don't return an `Error` here
```suggestion
        }).catch((err) => done(new Error('should not be hit: ' + JSON.stringify(err))))
```

test/unit/clients/connector_client_apple_authentication_tests.js
```diff
@@ -0,0 +1,132 @@
...
+
+    it('should return authorisation declined', function (done) {
+      connectorClient({ baseUrl: BASEURL }).chargeAuthWithWallet({
+        chargeId: TEST_CHARGE_ID,
+        provider: 'apple',
+        payload: appleAuthRequest.getPlain()
+      }).then(res => {
+        expect(res.body.message).to.be.equal('This transaction was declined.')
+        done()
+      }).catch((err) => done('should not be hit: ' + JSON.stringify(err)))
```
sfount:
```suggestion
      }).catch((err) => done(new Error('should not be hit: ' + JSON.stringify(err))))
```

test/unit/clients/connector_client_apple_authentication_tests.js
```diff
@@ -0,0 +1,132 @@
...
+
+    it('should return authorisation declined', function (done) {
+      connectorClient({ baseUrl: BASEURL }).chargeAuthWithWallet({
+        chargeId: TEST_CHARGE_ID,
+        provider: 'apple',
+        payload: appleAuthRequest.getPlain()
+      }).then(res => {
+        expect(res.body.message).to.be.equal('This transaction could be not be processed.')
+        done()
+      }).catch((err) => done('should not be hit: ' + JSON.stringify(err)))
```
sfount:
```suggestion
      }).catch((err) => done(new Error('should not be hit: ' + JSON.stringify(err))))
```

PR #752, opened 2019-03-21
PP-5015 - add new custom translation function

jonheslop:
i18n’s built in translation function `__()` returns the key
if it can’t find a translation. This is problematic as we show
something gross like 'cardDetails.title' to the user.
So this is a custom function that will check that the string
doesn’t equal the key. And if it does throw an error so the
template fails to render but with a nice helpful error message.

The error looks like this:
Screen Shot 2019-03-21 at 16 05 52

Which is how all the other Nunjucks errors look and it would mean you could never get your PR through tests

### DIFF
--- a/config/pay-translation.js
+++ b/config/pay-translation.js
@@ -1,0 +1,21 @@
+'use strict'
+// i18n’s built in translation function `__()` returns the key
+// if it can’t find a translation. This is problematic as we show
+// something gross like 'cardDetails.title' to the user.
+// So this is a custom function that will check that the string
+// doesn’t equal the key. And if it does throw an error so the
+// template fails to render but with a nice helpful error message.
+
+// NPM dependencies
+const i18n = require('i18n')
+
+module.exports = function (req, res, next) {
+  res.locals.__p = res.__p = function () {
+    const translation = i18n.__.apply(req, arguments)
+    if (arguments[0] === translation) {
+      throw new Error(`Template is missing a translation with ID: ${translation}`)
+    }
+    return translation
+  }
+  next()
+}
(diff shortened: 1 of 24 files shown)


jonheslop:
Instead of duplicating this PR in pay-products-ui I might move all of i18n and it’s config into pay-js-commons so that we have the config in one place

config/pay-translation.js
```diff
@@ -0,0 +1,21 @@
...
+// if it can’t find a translation. This is problematic as we show
+// something gross like 'cardDetails.title' to the user.
+// So this is a custom function that will check that the string
+// doesn’t equal the key. And if it does throw an error so the
+// template fails to render but with a nice helpful error message.
+
+// NPM dependencies
+const i18n = require('i18n')
+
+module.exports = function (req, res, next) {
```
sfount: This function should be named:
* Developers trying to understand this code 
* Stack trace

config/pay-translation.js
```diff
@@ -0,0 +1,21 @@
...
+  res.locals.__p = res.__p = function () {
+    const translation = i18n.__.apply(req, arguments)
+    if (arguments[0] === translation) {
+      throw new Error(`Template is missing a translation with ID: ${translation}`)
```
sfount: Depending on the behaviour we need - ideally we should not be throwing this error from middleware. The `next` middleware model designed to handle errors that occur during the express stack. 

This would look something like: 
```js 
const keyNotTranslated = arguments[0] === translation

if (keyNotTranslated) { 
  // no further middleware or controllers will be visited - this ends the flow
  next(new Error(`Template is missing a translation with ID: ${transaction}`))
  return
}

// everything is okay - continue processing the express stack
next()
```

config/pay-translation.js
jonheslop: 👍

config/pay-translation.js
jonheslop: I am not sure I understand this? This is part of the middleware stack?

PR #762, opened 2019-03-27
PP-4917 - Only Apple Pay v4 supports Maestro

jonheslop:
So check if browser supports v4 before enabling Apple Pay

### DIFF
--- a/app/views/charge.njk
+++ b/app/views/charge.njk
@@ -12,7 +12,7 @@
   {% block bodyStart %}
     {% if allowApplePay %}
       <script>
-        if (window.ApplePaySession && window.ApplePaySession.canMakePayments()) {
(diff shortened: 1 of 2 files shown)


app/views/charge.njk
```diff
@@ -12,7 +12,7 @@
   {% block bodyStart %}
     {% if allowApplePay %}
       <script>
-        if (window.ApplePaySession && window.ApplePaySession.canMakePayments()) {
+        if (window.ApplePaySession && window.ApplePaySession.canMakePayments() && window.ApplePaySession.supportsVersion(4)) {
```
alexbishop1: Don’t need so many brackets

alexbishop1: Sorry, misread it

PR #764, opened 2019-03-27
PP-4917 - Validate users email to check its legit

jonheslop:
Apple let users put anything in the email field, this makes sure its
legit.

I cant try this locally so I have no idea if it will work (and the docs
are rubbish) so would like to merge and try on test to see if it works.

Hence why I have left a log line because i imagine this wont work and
this will help me debug

### DIFF
--- a/app/assets/javascripts/browsered/web-payments/apple-pay.js
+++ b/app/assets/javascripts/browsered/web-payments/apple-pay.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const { prepareAppleRequestObject, showErrorSummary, toggleWaiting } = require('./helpers')
+const rfc822Validator = require('rfc822-validate')
 
 module.exports = () => {
   const session = new ApplePaySession(4, prepareAppleRequestObject())
@@ -30,7 +31,6 @@ module.exports = () => {
       .then(response => {
         session.completeMerchantValidation(response)
       }).catch(err => {
-        session.abort()
         showErrorSummary(i18n.fieldErrors.webPayments.apple)
         ga('send', 'event', 'Apple Pay', 'Error', 'Error completing Merchant validation')
         return err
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/browsered/web-payments/apple-pay.js
```diff
@@ -30,7 +31,6 @@ module.exports = () => {
       .then(response => {
         session.completeMerchantValidation(response)
       }).catch(err => {
-        session.abort()
```
oswaldquek: why does this line need to be removed?

app/assets/javascripts/browsered/web-payments/apple-pay.js
```diff
@@ -70,5 +70,16 @@ module.exports = () => {
     })
   }
 
+  session.onshippingcontactselected = function (event) {
```
oswaldquek: 1. is this method called by magic?
2. does `onshippingcontactselected` need to all be in lower case?

app/assets/javascripts/browsered/web-payments/apple-pay.js
```diff
@@ -70,5 +70,16 @@ module.exports = () => {
...
+    console.log(event)
+    const { email } = event.shippingContact
+
+    if (rfc822Validator(email)) {
```
gidsg: could you use a ternary operator for this if statement?

app/assets/javascripts/browsered/web-payments/apple-pay.js
jonheslop: Because it raised an error, when I run locally (where I wont validate as a merchant) and this got called in caused an error so the JS and so the error message below doesn’t show. Basically it’s too early to call abort here, becuase a valid Apple Pay session hasn’t begun

app/assets/javascripts/browsered/web-payments/apple-pay.js
jonheslop: Yeah https://developer.apple.com/documentation/apple_pay_on_the_web/applepaysession/1778008-completeshippingcontactselection

app/assets/javascripts/browsered/web-payments/apple-pay.js
jonheslop: Yeah good point

PR #767, opened 2019-03-28
PP-5023: Emit time to expiry metric

oswaldquek:
Emit the Apple Pay merchant identity certificate's time-to-expiry (in
milliseconds) metric on server startup. 

There are few `logger.debug` lines which I will delete when I've verified it all works in the test environment.

## WHAT

## HOW 
_Steps to test or reproduce:_

### DIFF
--- a/app/utils/metrics.js
+++ b/app/utils/metrics.js
@@ -1,13 +1,45 @@
 'use strict'
 
+const net = require('net')
+const logger = require('winston')
+const certinfo = require('cert-info')
 const appmetrics = require('appmetrics')
 const metricsHost = process.env.METRICS_HOST || 'localhost'
 const metricsPort = process.env.METRICS_PORT || 8125
 const metricsPrefix = 'frontend.'
+const { APPLE_PAY_MERCHANT_ID_CERTIFICATE } = process.env
 
-exports.metrics = () => {
+const sendInitialCertificateMetric = function sendInitialCertificateMetric () {
+  if (APPLE_PAY_MERCHANT_ID_CERTIFICATE !== undefined) {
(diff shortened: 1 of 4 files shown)


app/utils/metrics.js
```diff
@@ -1,13 +1,45 @@
...
 const appmetrics = require('appmetrics')
 const metricsHost = process.env.METRICS_HOST || 'localhost'
 const metricsPort = process.env.METRICS_PORT || 8125
 const metricsPrefix = 'frontend.'
+const { APPLE_PAY_MERCHANT_ID_CERTIFICATE } = process.env
 
-exports.metrics = () => {
+const sendInitialCertificateMetric = function sendInitialCertificateMetric () {
+  if (APPLE_PAY_MERCHANT_ID_CERTIFICATE !== undefined) {
+    var merchantIdCert = certinfo.info(APPLE_PAY_MERCHANT_ID_CERTIFICATE)
```
sfount: We shouldn't be mixing `var` and `const`. `const` should be used here

app/utils/metrics.js
```diff
@@ -1,13 +1,45 @@
...
+    const certificateKey = 'apple-pay.merchant-identity-certificate.time-to-expiry'
+    try {
+      const socket = net.createConnection(metricsPort, metricsHost, () => {
+        logger.debug(`Socket opened with ${metricsHost} ${metricsPort}`)
+        socket.write(`${metricsPrefix}.${certificateKey} ${certificateTimeToExpiry}\n`)
+        socket.end()
+      })
+
+      socket.on('data', (data) => {
+        logger.debug('Socket said', data)
```
sfount: If this will be kept in production code it should say give more information. Maybe something along the lines of 'Data event from certificate expiration socket:'

oswaldquek:
@sfount fixed

PR #769, opened 2019-03-28
PP-4917 - Validate email after payment authorised

jonheslop:
I tried to validate the email before payment authorisation but Apple
won’t hand over any personally identifiable information before
authorisation, will only give things like contry and postcode.

So we have to do this after authorisation. I have refactored the test so
that we can construct an Apple Pay Session that returns a bad email

Have written test to confirm this behaviour

### DIFF
--- a/test/cypress/integration/web-payments/apple-pay.spec.js
+++ b/test/cypress/integration/web-payments/apple-pay.spec.js
@@ -3,31 +3,33 @@ describe('Apple Pay payment flow', () => {
   const chargeId = 'ub8de8r5mh4pb49rgm1ismaqfv'
   const returnURL = '?success'
 
-  const validPaymentRequestResponse = {
-    shippingContact: {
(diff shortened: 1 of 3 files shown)


test/cypress/integration/web-payments/apple-pay.spec.js
```diff
@@ -196,5 +204,42 @@ describe('Apple Pay payment flow', () => {
       cy.get('#enter-card-details-container').should('be.visible')
       cy.get('#card-no').should('be.visible')
     })
+
+    it('Should setup the payment and load the page', () => {
```
oswaldquek: why is this test necessary?

jonheslop: Because between each test we need to reset the view

jonheslop: I guess this makes it clear, it would be weird not to describe what a part of the test is doing

PR #780, opened 2019-04-02
PP-4917 add more digital wallet logging

jonheslop:
So we can better understand how digital wallet payments are performing
we are adding logging lines so that we can tell what they were presented
with and what they chose.

### DIFF
--- a/app/assets/javascripts/browsered/web-payments/index.js
+++ b/app/assets/javascripts/browsered/web-payments/index.js
@@ -12,6 +12,7 @@ const standardMethodContainer = document.getElementById('enter-card-details-cont
 const initApplePayIfAvailable = () => {
   if (document.body.classList.contains('apple-pay-available')) {
     document.getElementById('payment-method-apple-pay').checked = true
+    ga('send', 'event', 'Apple Pay', 'Enabled', 'Apple pay available on this device')
   }
 }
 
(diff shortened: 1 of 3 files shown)


oswaldquek:
Can the commit https://github.com/alphagov/pay-frontend/pull/780/commits/04649573f7784dd5723cc31f570ebdae20c0d31e follow the commit format please. E.g. "PP-4917: Log what user is presented with...". I'd also like to see a commit message explaining why we're adding these analytics (i.e. it's for the funnel to see the conversion rate of users who choose webpayments).

app/assets/javascripts/browsered/web-payments/index.js
```diff
@@ -9,9 +9,16 @@ const { createGooglePaymentRequest, googlePayNow } = require('./google-pay')
 const paymentMethodForm = document.getElementById('web-payments-container')
 const standardMethodContainer = document.getElementById('enter-card-details-container')
 
+const titleCase = function stringToTitleCase(string) {
+  return string.replace(/\w\S*/g, txt => {
+    return txt.charAt(0).toUpperCase() + txt.substr(1).toLowerCase()
+  })
```
oswaldquek: what does this function do?

jonheslop: the form submission has values like `apple-payz so it make it into `Apple Pay` which means that all our google analytics categories are matching

PR #781, opened 2019-04-02
PP-5057 do not post to card_id when card no is less than 11 digits

SandorArpa:
## WHAT
- This change stops posting for a card check to card id when the card number is less than 11 digits thus stopping card_id throwing RuntimeException which result in a 500 error.

with @jonheslop

### DIFF
--- a/app/assets/javascripts/modules/form-card-type.js
+++ b/app/assets/javascripts/modules/form-card-type.js
@@ -139,8 +139,7 @@ var showCardType = function() {
     return new Promise(function (resolve, reject) {
       var request = new XMLHttpRequest();
       var card = getCardType();
-      // this should already be picked up by the other validations
-      if (card.length !== 1) return
+      if (card.length !== 1 && cardInput.value.replace(/\D/g,'').length < 11) return
       request.open('POST', '/check_card/' + chargeId, true);
       request.setRequestHeader('Content-type', 'application/json');
       request.onload = function () {
--- a/app/models/card.js
+++ b/app/models/card.js
@@ -31,6 +31,10 @@ const checkCard = function (cardNo, allowed, language, correlationId, subSegment
     }
 
     AWSXRay.captureAsyncFunc('cardIdClient_post', function (postSubsegment) {
+      if (cardNo.length > 0 && cardNo.length < 11) {
+        postSubsegment.close()
+        return reject(new Error(i18n.__('fieldErrors.fields.cardNo.numberIncorrectLength')))
+      }
       cardIdClient.post({ payload: data, correlationId: correlationId }, postSubsegment)
         .then((response) => {
           postSubsegment.close()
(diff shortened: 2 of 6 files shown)


app/assets/javascripts/modules/form-card-type.js
```diff
@@ -135,12 +135,12 @@ var showCardType = function() {
     };
   }
 
-  var checkCardtypeIsAllowed = function () {
+      var checkCardtypeIsAllowed = function () {
     return new Promise(function (resolve, reject) {
       var request = new XMLHttpRequest();
       var card = getCardType();
       // this should already be picked up by the other validations
```
alexbishop1: This comment is out of date now because nowhere prior to this does the length 11 check

app/controllers/charge_controller.js
```diff
@@ -183,9 +183,13 @@ module.exports = {
   checkCard: (req, res) => {
     const namespace = getNamespace(clsXrayConfig.nameSpaceName)
     const clsSegment = namespace.get(clsXrayConfig.segmentKeyName)
+    const normalisedCardNo = normalise.creditCard(req.body.cardNo)
+    if (normalisedCardNo && normalisedCardNo.length < 11) {
+      return res.json({ 'accepted': false, message: 'Card number is not the correct length' })
+    }
```
alexbishop1: I think it might be better to have this check inside the `checkCard` function in `card.js` and have a `return reject(new Error('Card number is not the correct length'))` in there — that way the existing code in this file for constructing the response would just work

test/cypress/integration/card/payment.spec.js
```diff
@@ -148,4 +148,32 @@ describe('Standard card payment flow', () => {
...
+      cy.task('setupStubs', checkCardDetailsStubs)
+
+      cy.server()
+      cy.route('POST', `/check_card/${chargeId}`).as('checkCard')
+
+      cy.get('#card-no').type('1234567890')
+      cy.get('#card-no').blur()
+
+      cy.get('#card-no').should('have.class', 'govuk-input--error')
+      cy.get('#card-no-lbl').should('contain', 'Card number is not the correct length')
```
alexbishop1: Ah, I didn’t realise this could actually be displayed on the page. We’d need a Welsh translation. Might be easier to reuse whatever the current message is that we have for a card not being supported.

app/assets/javascripts/modules/form-card-type.js
```diff
@@ -135,12 +135,11 @@ var showCardType = function() {
     };
   }
 
-  var checkCardtypeIsAllowed = function () {
+      var checkCardtypeIsAllowed = function () {
```
alexbishop1: Was this indentation intentional?

app/models/card.js
```diff
@@ -31,6 +31,10 @@ const checkCard = function (cardNo, allowed, language, correlationId, subSegment
     }
 
     AWSXRay.captureAsyncFunc('cardIdClient_post', function (postSubsegment) {
+      if (cardNo.length > 0 && cardNo.length < 11) {
+        postSubsegment.close()
+        return reject(new Error(i18n.__('fieldErrors.fields.cardNo.numberIncorrectLength')))
```
alexbishop1: I see, we already have an i18n string for this. Awesome!

PR #782, opened 2019-04-02
BAU - Remove Cypress dev dependency

jonheslop:
Cypress takes a significant amount of time to both download and unzip in
CI/ build processes - it should only be used via pre-compiled containers
during the build pipeline so having the developer set up their own
envrionment is the lesser of two evils.

A copy of — https://github.com/alphagov/pay-toolbox/pull/53


heathd:
closing since it's been open for 2 months

PR #783, opened 2019-04-02
PP-5062 - Make sure corrected email gets sent to connector

jonheslop:
The patch which sends the email was moved above the code that corrected
typos, meaning the uncorrected email was sent.

### DIFF
--- a/test/cypress/integration/card/email-typo.spec.js
+++ b/test/cypress/integration/card/email-typo.spec.js
@@ -1,0 +1,238 @@
+const cardPaymentStubs = require('../../utils/card-payment-stubs')
+
+describe('Standard card payment flow', () => {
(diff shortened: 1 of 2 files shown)


test/cypress/integration/card/email-typo.spec.js
```diff
@@ -0,0 +1,321 @@
+const cardPaymentStubs = require('../../utils/card-payment-stubs')
+const lodash = require('lodash')
+
+describe('Standard card payment flow', () => {
+  const tokenId = 'be88a908-3b99-4254-9807-c855d53f6b2b'
+  const chargeId = 'ub8de8r5mh4pb49rgm1ismaqfv'
+
+  const invalidEmail = 'validpayingemail@gnail.com'
```
alexbishop1: Not really sure about the term `invalidEmail` because it might be valid. How about `likelyTypoEmail` or something?

PR #793, opened 2019-04-08
PP-5072 more digital wallet payment tests

jonheslop:
After review increasing test coverage

### DIFF
--- a/test/cypress/integration/card/payment.spec.js
+++ b/test/cypress/integration/card/payment.spec.js
@@ -1,5 +1,28 @@
 const cardPaymentStubs = require('../../utils/card-payment-stubs')
 const { getMockApplePayClass } = require('../../utils/apple-pay-js-api-stubs')
+const { getMockPaymentRequest } = require('../../utils/payment-request-api-stub')
+
+const validPaymentRequestResponse = {
+  details: {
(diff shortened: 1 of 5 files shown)


test/cypress/integration/card/payment.spec.js
```diff
@@ -205,4 +229,127 @@ describe('Standard card payment flow', () => {
...
+          // Stub Payment Request API
+          if (win.PaymentRequest) {
+            // If we’re running in headed mode
+            cy.stub(win, 'PaymentRequest', getMockPaymentRequest(validPaymentRequestResponse))
+          } else {
+            // else headless
+            win.PaymentRequest = getMockPaymentRequest(validPaymentRequestResponse)
+          }
+        }
+      })
```
oswaldquek: if the service is not Apple Pay enable why do we need to `getMockPaymentRequest`? `getMockPaymentRequest` looks like a response from Google Pay

jonheslop: 🤦‍♂️ I’ve got the labels backwards, it should say Google Pay (and the Apple Pay one should say Google Pay). Will switch now

The point in this line is to add payment request API support to our cypress testing browser

PR #804, opened 2019-04-15
BAU - Generate static assets after install

jonheslop:
Sometimes people forget to do this which causes issues when running
locally. I added the script to cypress:server but this penalises people
who run this often so moving to prepare

Tried it out by running `rm -rf node_modules && npm i` and can confirm it works

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     ]
   },
   "scripts": {
+    "prepare": "npm run compile",
     "compile": "grunt generate-assets",
     "clean": "grunt clean",
     "start": "node start.js",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -31,14 +31,15 @@
     ]
   },
   "scripts": {
+    "postinstall": "npm run compile",
```
sfount: You can use the `prepare` hook here and it will only run if dev dependencies are installed. As compile relies on these dependencies this will prevent errors installing only for production. 

(It would be good to double check this works as expected before committing to it)

jonheslop: Ok moved it to prepare also did `rm -rf node_modules && npm i` and confirm it works

PR #841, opened 2019-05-03
BAU (pending ticket): Fix async Google pay check

sfount:
Google Pay async check :: 

* Currently checks if(Promise) -- this will always be true 
* Update to use Promise.then

Before merging/ testing: 
* This will introduce a wait before we know if Google Pay is available -- the user experience should be validated


sfount:
Note this should be reverted and superseded by https://github.com/alphagov/pay-frontend/pull/842. Merging now to verify behaviour in staging.

PR #842, opened 2019-05-03
PP-5176: Fix client side async Google pay check

sfount:
Google Pay async check ::

* Currently checks if(Promise) -- this will always be true
* Update to use Promise.then

Before merging/ testing:

This will introduce a wait before we know if Google Pay is available -- the user experience should be validated

Supersedes https://github.com/alphagov/pay-frontend/pull/841


sfount:
This functionality was merged in https://github.com/alphagov/pay-frontend/pull/841

PR #844, opened 2019-05-07
PP-5539: Set json formatting for app logs

paroxp:
## WHAT

Current version is slightly out of date. We need to be able to set the
JSON structured logging when in `production` environment, which with
version three will be more pragmatic to do.

At the moment, different files and components are loading `winston`
directly, which means the setup would have to be duplicated in many
places or potentially diverge.

This isn't cool. Adding a function that will be loaded from everywhere
instead, gives us the lever of abstraction we need without refactoring
the app to use context.

## HOW 

- Sanity check
- Run tests
- Perhaps test it in actual env running?


paroxp:
I was bored waiting for other things to complete. Sorry.

config/logger.js
```diff
@@ -0,0 +1,13 @@
+'use strict'
+
+const winston = require('winston')
+
+module.exports.setupLogger = function () {
+  return winston.createLogger({
```
sfount: This could take advantage of the singleton pattern with Node modules where the module is initialised once and then shared by all other modules -- what is happening here is a new instance (and object to track/ store in memory) is being created by every file that requires logging. 

Suggestion -- 

```js
const winston = require('winston')

const config = {} //...
const logger = winston.createLogger(config)

module.exports = logger
```

```js 
const logger = require('../../config/logger')
logger.info()
```

In this scenario the (relatively large) instance is only created once and is then used by all modules.

config/logger.js
```diff
@@ -0,0 +1,13 @@
...
+    level: process.env.NODE_ENV !== 'production' ? winston.level.debug : winston.level.info,
```
sfount: This is a great start, as it gets more complex it may be useful to explicitly define separate production and development configurations and have the only branch be which one is used. 

Something like: 
```js
const developmentConfig = {} //...
const productionConfig = {} //... 
```

config/logger.js
paroxp: I agree. I'd much rather introduce a shared context of some sort, but decided to go the way it use to be in this project 🤔 

I'll make that change if you're happy with it. 🎉

sfount:
This was extended and merged in https://github.com/alphagov/pay-frontend/pull/1108

PR #861, opened 2019-05-13
PP-5205 - Nest spans so the clauses can sit together without CSS

jonheslop:
We have 3 scenearios here:
- Default - 3 digit CVC hint
- AmEx - 4 digit CVC hint
- No-js - both hints togother. With 'For American Express' clause to make it read better.

These spans were all siblings and we're smushed together with floats and margins, this is convoluted and cause the flow to break because the CVC input floated to the end.

We can do it more simlpy by nesting them and using `display:inline`. But because HTML will automatically put a space if there is a whitespace between spans, we have to smush it together in the markup so it‘s all one line.

## Before

Screen Shot 2019-05-13 at 14 38 49

## After

Screen Shot 2019-05-13 at 14 53 56

### DIFF
--- a/app/assets/sass/modules/_cvc.scss
+++ b/app/assets/sass/modules/_cvc.scss
@@ -2,43 +2,40 @@
   display: none;
 }
 
-body:not(.js-enabled) .govuk-form-group.cvc .hidden {
-  display: block;
-  visibility: visible;
-  margin-right: 3px;
-  float: left;
-}
+body:not(.js-enabled) {
+  .generic-cvc {
+    display: block;
+  }
 
-body:not(.js-enabled) .govuk-form-group.cvc .amex-cvc {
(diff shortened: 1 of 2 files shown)


locales/cy.json
```diff
@@ -17,7 +17,7 @@
 		"cvc": "Cod diogelwch y cerdyn",
 		"cvcTip": "Y 3 digid olaf ar gefn y cerdyn",
 		"amexcvcTip": "Y 4 digid olaf ar flaen y cerdyn ar ôl rhif y cerdyn",
-		"amexcvcNonjs": "Am American Express,",
+		"amexcvcNonjs": ". Am American Express,",
```
alexbishop1: Makes this harder to reuse in other places but I suppose we can live with that

PR #872, opened 2019-05-16
PP-5226 - typographic heirarchy improvements

jonheslop:
## Before 
Screenshot_2019-05-16 Enter card details(1)

## After
Screenshot_2019-05-16 Enter card details(2)


alexbishop1:
Gap between the _Contact details_ heading and the email field looks a bit large.

jonheslop:
@alexbishop1 Made gap consistent and added in borders too because otherwise the larger gaps make it look floaty

PR #873, opened 2019-05-16
PP-5228 - update design of summary box

jonheslop:
## Before
Screen Shot 2019-05-16 at 14 51 20

## After 
Screen Shot 2019-05-16 at 14 50 57

### DIFF
--- a/app/assets/sass/modules/_summary-panel.scss
+++ b/app/assets/sass/modules/_summary-panel.scss
@@ -1,6 +1,6 @@
 .payment-summary {
   margin-top: govuk-spacing(6);
-  border: govuk-spacing(2) solid $govuk-border-colour;
+  border-top: 2px solid $govuk-brand-colour;
   padding: govuk-spacing(3);
   background-color: govuk-colour('grey-4');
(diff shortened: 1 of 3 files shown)


alexbishop1:
What’s it look like on mobile?

jonheslop:
The same, it didn’t change on mobile before and it still doesn’t. 

Conor and I are making small adjustments one at a time. We’re aware it could be better on mobile. But that’s mainly about positioning not style

app/assets/sass/modules/_summary-panel.scss
```diff
@@ -1,6 +1,6 @@
 .payment-summary {
   margin-top: govuk-spacing(6);
-  border: govuk-spacing(2) solid $govuk-border-colour;
+  border-top: 2px solid $govuk-brand-colour;
```
danworth: I guess using `govuk-spacing()` offered some uniformity and means to globally adjust spacing consistently. But if it doesn't offer an actual spacing of 2px and that is what is required then 🤷‍♂

PR #874, opened 2019-05-16
PP-5229 - New design for showing digital wallet options

jonheslop:
This simplifies it somewhat, we don’t hide the default payment form we just show a google/apple button above it.

- rewrote markup for new design
- added logos and css to style it all
-updated client side JS to work with changed page elements
- Updated tests to look for and interact with new elements

Changes are easier if you turn whitespace diff off - https://github.com/alphagov/pay-frontend/pull/874/files?w=1 (it makes it `+65 -162` as well 😄)

01 enter-card-details


test/cypress/integration/web-payments/google-pay.spec.js
```diff
@@ -130,50 +128,6 @@ describe('Google Pay payment flow', () => {
       cy.location('pathname').should('eq', `/card_details/${chargeId}`)
     })
 
-    it('Should show Google Pay as a payment option but user chooses to pay normally', () => {
```
jonheslop: This test was in there twice?!

jonheslop:
At the moment there is no case where they could both show at once, this could change if Safari supported google pay. 

Yeah gonna merge and then test manually on `test-12`

PR #881, opened 2019-05-21
PP-5230 update confirmation page design

jonheslop:
- No more payment summary floating on the right, its all in tables
- Bigger clearer pay button at the top

02 confirm-your-payment


alexbishop1:
Is it unchanged on mobile?

jonheslop:
It looks like this

Screenshot_2019-05-21 Confirm your payment

alexbishop1:
In the past, we deliberately changed the confirm page so that it had the grey summary box in the same position as the card details page. This reduced the number of people who didn’t complete their payment after reaching the confirm page. This design changes it back somewhat. Are we assuming that the more prominent confirm button will counteract the reduced visual continuity between the card details page and the confirm page?

jonheslop:
I think we’re just gonna try it out and see how conversion changes.

The summary was moved to the bottom on mobile which sorta makes it redundant and on desktop, things aligned to the right cause issues for people using zoom text or alike.

The idea is that having the button bigger and at the top makes it clear, the journey  isn’t over and users need to click it to finish. We’re hoping it decreases the number of abandoned payments

PR #889, opened 2019-05-23
PP-5279 - fix Apple Pay when Google Pay is also availabled

jonheslop:
If both are available we had overrode the Apple Button with a Google one.

Now we make a separate form for each and use Javascript/CSS to determine which one to present to user.

Had to also rejig the JS to understand there are two forms on the page and it needs to set event listeners on both

### DIFF
--- a/app/views/charge.njk
+++ b/app/views/charge.njk
@@ -56,32 +56,39 @@
         </ul>
       </div>
     </div>
-    {% if allowApplePay or allowGooglePay %}
-      {% if allowApplePay %}
-        {% set buttonHTML %}
-            {{ __p("cardDetails.webPayments.payWith") }}
-            <span class="govuk-visually-hidden">Apple Pay</span>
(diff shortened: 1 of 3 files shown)


app/views/charge.njk
```diff
@@ -56,32 +56,39 @@
...
-        {% set buttonHTML %}
-            {{ __p("cardDetails.webPayments.payWith") }}
-            <span class="govuk-visually-hidden">Apple Pay</span>
-            <img role="presentation" class="web-payments-logo" src="/images/apple-pay-logo.svg"/>
-        {% endset %}
-        {% set webPaymentValue = 'Apple Pay'%}
-        {% set webPaymentClass = 'apple-pay'%}
-      {% endif %}
-      {% if allowGooglePay %}
-        {% set buttonHTML %}
```
jonheslop: This was the problem, this set overrode the one above it

PR #890, opened 2019-05-23
PP-5279 - add tests to check web payments work when both enabled for …

jonheslop:
…a service

Tests to go with #889


jonheslop:
Messed this up closing and starting again (pushed a revert that i did just to test it would have flagged the issue)

PR #892, opened 2019-05-23
RED-46: Stop using our node base image

rhowe-gds:
Our base image doesn't do much. Import what it does into the local Dockerfile


Dockerfile
```diff
@@ -1,5 +1,15 @@
-# govukpay/nodejs:alpine-3.8.1
-FROM govukpay/nodejs@sha256:a3ae1b0b3ab5e7cdefe0787b97b1047cdc5e0c5166fb677e3cefac538bc99f86
+# alpine:3.8
+FROM alpine@sha256:ea47a59a33f41270c02c8c7764e581787cf5b734ab10d27e876e62369a864459
```
sfount: Is there an advantage to using alpine + add `npm`, `nodejs` over the nodejs alpine images? 
https://github.com/nodejs/docker-node/blob/5a6a5e91999358c5b04fddd6c22a9a4eb0bf3fbf/10/alpine/Dockerfile

Dockerfile
```diff
@@ -1,5 +1,15 @@
-# govukpay/nodejs:alpine-3.8.1
-FROM govukpay/nodejs@sha256:a3ae1b0b3ab5e7cdefe0787b97b1047cdc5e0c5166fb677e3cefac538bc99f86
+# alpine:3.8
```
sfount: The current base image pull from alpine 3.9 -- just making sure this regression is intentional

Dockerfile
rhowe-gds: I don't know - this is directly equivalent to what we do today, just importing the code from the current base image `Dockerfile` in `pay-infra`

Dockerfile
rhowe-gds: The current code on `pay-infra` `master` does, but that's not the code which generated `govukpay/nodejs:alpine-3.8.1`

Alpine 3.9 includes node 10 and appmetrics fails to `npm install` on that for some reason I haven't got to the bottom of yet

PR #893, opened 2019-05-24
BAU Release product page with fixed link

stephencdaly:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1781,7 +1781,7 @@
     },
     "ansi-escapes": {
       "version": "1.4.0",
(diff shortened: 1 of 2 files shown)


alexbishop1:
Any idea why this changes lots of `resolved` entries in the package-lock.json from HTTPS to HTTP?

stephencdaly:
Node rubbish, Jon's said its fine before. Would be ideal if that didn't happen but shouldn't be a problem

alexbishop1:
According to https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285/47 it’s a registry bug but any actual downloads will always use HTTPS so it’s basically just cosmetic.

stephencdaly:
ah good to know, thanks 👍

PR #921, opened 2019-06-25
PP-5366 - use inputmode and pattern attribute on numeric fields

jonheslop:
The Design System team have done some accessibility testing on `input
type="number"` and deemed them problematic. They have suggested using

`input type="text" inputmode="numeric" pattern="[0-9]*"`

instead as documented here https://github.com/alphagov/govuk-frontend/issues/1449


alexbishop1:
I forgot to say — it would be good to check if this changes anything to do with browser credit card autofill or that fancy feature where iOS Safari offers to read the card number from your physical card using the camera.

jonheslop:
> I forgot to say — it would be good to check if this changes anything to do with browser credit card autofill or that fancy feature where iOS Safari offers to read the card number from your physical card using the camera.

Yeah will check this, I think that should be fine as I think that mostly relies on the `autocomplete` attribute values

PR #952, opened 2019-07-16
PP-5560 - govuk-frontend 3.0 upgrade

jonheslop:
Updated all templates to meet breaking changes as part of 3.0.0 release which are all detailed here — https://github.com/alphagov/govuk-frontend/releases/tag/v3.0.0

This 3.0 release includes numerous accessibility updates but most visually prominent is the all new colour scheme - https://designnotes.blog.gov.uk/2019/07/29/weve-made-the-gov-uk-design-system-more-accessible/

### DIFF
--- a/app/assets/sass/modules/_button-reset.scss
+++ b/app/assets/sass/modules/_button-reset.scss
@@ -15,6 +15,20 @@
   &:hover {
     color: $govuk-link-hover-colour;
   }
+
+  &:focus {
+    @include govuk-not-ie8 {
+      outline: $govuk-focus-width solid transparent;
+      outline-offset: 0;
+    }
+    color: $govuk-text-colour;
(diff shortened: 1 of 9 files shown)


jonheslop:
Probs need to do some comms before this release or at least check with Katie

app/assets/sass/modules/_button-reset.scss
```diff
@@ -15,6 +15,20 @@
   &:hover {
     color: $govuk-link-hover-colour;
   }
+
+  &:focus {
```
nickcolley: Could this be using the mixin?

The box-shadow's look like a pre-release version of the focus.

https://design-system.service.gov.uk/get-started/focus-states/

jonheslop: Ah nice one, will update

PR #970, opened 2019-08-20
PP-5546 Add client method to get JWT from connector

stephencdaly:
Add a method to connector client to call the connector endpoint to get a Worldpay 3DS Flex DDC JWT.

Add a pact test for this interaction.


app/services/clients/connector_client.js
```diff
@@ -53,6 +54,9 @@ const _getDeleteTokenUrlFor = tokenId => baseUrl + CARD_DELETE_CHARGE_TOKEN_PATH
 /** @private */
 const _getPatchUrlFor = chargeId => baseUrl + CARD_CHARGE_PATH.replace('{chargeId}', chargeId)
 
+/** @private */
```
oswaldquek: Does this line do anything?

stephencdaly: i doubt it, i don't really know. doesn't look like we use this notation anywhere else - not sure why it's done here

oswaldquek: ask @jonheslop !

stephencdaly: he doesn't know. i may remove it all in another pr if i get the chance

PR #971, opened 2019-08-20
PP-5546 Get Worldpay 3DS Flex JWT on card details page

stephencdaly:
If a gateway account has payment provider "worldpay" and 3DS integration version of 2, call connector to get the JWT for DDC and store it on the window.Charge object.

Have added unit tests for worldpay_3ds_flex_service.js, but not integration tests as appropriate integration tests should be added when the device data collection is implemented.


stephencdaly:
Pushed another commit to handle non-200 status codes from connector

PR #972, opened 2019-08-21
PP-5545: Pact test for authorising card

oswaldquek:
This test is really to ensure `worldpay_3ds_flex_ddc_result` is a valid
property in the auth request to connector.

## WHAT

## HOW 
_Steps to test or reproduce:_

### DIFF
--- a/test/unit/clients/connector_client_charge_tests.js
+++ b/test/unit/clients/connector_client_charge_tests.js
@@ -12,6 +12,7 @@ const { PactInteractionBuilder } = require('../../fixtures/pact_interaction_buil
 
 const TEST_CHARGE_ID = 'testChargeId'
 const GET_CHARGE_URL = `/v1/frontend/charges/${TEST_CHARGE_ID}`
+const AUTHORISE_CHARGE_URL = `/v1/frontend/charges/${TEST_CHARGE_ID}/cards`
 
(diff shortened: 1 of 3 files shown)


test/unit/clients/connector_client_charge_tests.js
```diff
@@ -30,30 +31,55 @@ describe('connector client - charge tests', function () {
...
 
+  describe('making auth request', function () {
+    const authRequest = fixtures.validAuthorisationRequest()
+    
+    before(() => {
+      const response = fixtures.validChargeCardDetailsAuthorised()
+      const builder = new PactInteractionBuilder(AUTHORISE_CHARGE_URL)
+        .withRequestBody(authRequest.getPactified())
+        .withMethod('POST')
+        .withState('a sandbox account exists with a charge with id testChargeId that is in state ENTERING_CARD_DETAILS.')
```
danworth: The `worldpay_3ds_flex_ddc_result` is only applicable to worldpay accounts, should that be reflected in the state rather than using `sandbox`?

oswaldquek: I did think about that but thought it wouldn't work to have worldpay, otherwise the provider test would actually try to call worldpay...

PR #983, opened 2019-09-09
PP-5620: Pact test for /v1/frontend/tokens/{tokenId} endpoint

oswaldquek:
## WHAT

## HOW 
_Steps to test or reproduce:_

### DIFF
--- a/test/fixtures/payment_fixtures.js
+++ b/test/fixtures/payment_fixtures.js
@@ -218,6 +218,23 @@ const buildChargeDetailsWithPrefilledCardHolderDeatils = (opts) => {
 }
 
 const fixtures = {
+  tokenResponse: (opts = {}) => {
+    const data = {
+      'used': opts.used,
+      'charge': {
+        'externalId': opts.chargeExternalId
+      }
+    }
+    return {
+      getPactified: () => {
+        return pactBase.pactify(data)
+      },
+      getPlain: () => {
+        return data
+      }
+    }
+  },
+
   // intitial charge details returned have different API surface than charge others
   validChargeCreatedByToken: (opts = {}) => {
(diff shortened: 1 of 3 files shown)


test/fixtures/payment_fixtures.js
```diff
@@ -218,6 +218,23 @@ const buildChargeDetailsWithPrefilledCardHolderDeatils = (opts) => {
 }
 
 const fixtures = {
+  tokenResponse: (opts = {}) => {
+    const data = {
+      'used': opts.used,
+      'charge': {
+        'externalId': opts.chargeExternalId
```
danworth: Should this be snake case `external_id`? Also, is `external_id` the only element in the charge response that is required for now and that's why the rest of the elements in `charge` aren't being tested?

oswaldquek: Yes the current `externalId` is correct as that's what the connector service responds with.
Also yes to the second question. If you have a read of the scenarios in https://payments-platform.atlassian.net/browse/PP-5620 we only care about the `used` indicator and payment id.

oswaldquek: That's why I asked about snake case and camel case in the slack channel earlier as I thought it was snake case originally. But when running the pact test it turned out to be camel.

danworth: Ah I get it now, in Connector `ChargeEntity` has no Jackson annotations so the element names match the camelCase format of the class fields, the `GatewayAccountEntity` has getters with Jackson annotations and their elements are snake_case, hence the `"gatewayAccount.service_name"`... so yes its a confusing.

Also, yes looking at the scenarios the only important bit is the external id 👍

PR #991, opened 2019-09-10
PP-5626 x gov analytics setup part 1

jonheslop:
Cleaning things up and using new ENV var rather than hardcoding in ID


app/views/includes/analytics.njk
```diff
@@ -20,9 +21,12 @@
     });
 
     {# X-Gov google analtyics #}
```
rhowe-gds:
```suggestion
    {# X-Gov google analytics #}
```

jonheslop:
Parking this as there would other work we would need to do to the API that we're not going right now

PR #992, opened 2019-09-10
PP-5620: Replace get token call

oswaldquek:
Replace calls to `/v1/frontend/tokens/TOKEN_ID/charge` with
`/v1/frontend/tokens/TOKEN_ID/`

## WHAT

## HOW 
_Steps to test or reproduce:_


alexbishop1:
Does this work without any other changes? The `/v1/frontend/tokens/TOKEN_ID/charge` returns a charge object whereas `/v1/frontend/tokens/TOKEN_ID/` returns an object containing a `used` property and a `charge` token. I would have thought we’d have to drill down into the `charge` property of the response.

oswaldquek:
@alexbishop1 yeah it doesn't work. e2e tests fail. thanks for the catch.

PR #1021, opened 2019-09-20
PP-5646: Extra information on system error logs

oswaldquek:
We've had some instances where the secure URL have been fetched more than once;
however we currently don't log the gateway account information which would be
useful to do some analysis on.

## WHAT

## HOW 
_Steps to test or reproduce:_

### DIFF
--- a/app/controllers/secure_controller.js
+++ b/app/controllers/secure_controller.js
@@ -20,11 +20,14 @@ const paths = require('../paths')
 exports.new = async function (req, res) {
   const chargeTokenId = req.params.chargeTokenId || req.body.chargeTokenId
   const correlationId = req.headers[CORRELATION_HEADER] || ''
+  var chargeId
+  var gatewayAccountId
+  var gatewayAccountType
   try {
     const chargeData = await Charge(correlationId).findByToken(chargeTokenId)
-    const chargeId = chargeData.charge.externalId
-    const gatewayAccountId = chargeData.charge.gatewayAccount.gateway_account_id
-    const gatewayAccountType = chargeData.charge.gatewayAccount.type
+    chargeId = chargeData.charge.externalId
+    gatewayAccountId = chargeData.charge.gatewayAccount.gateway_account_id
+    gatewayAccountType = chargeData.charge.gatewayAccount.type
     if (chargeData.used === true) {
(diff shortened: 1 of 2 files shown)


app/controllers/secure_controller.js
```diff
@@ -47,7 +50,7 @@ exports.new = async function (req, res) {
         'the frontend state cookie not existing, or the frontend state cookie containing an invalid value.')
       return responseRouter.response(req, res, 'UNAUTHORISED')
     }
-    logging.systemError('Secure controller token', correlationId, chargeTokenId)
+    logging.systemError('Secure controller token', correlationId, chargeId, gatewayAccountId, gatewayAccountType)
```
jankowiakmaria: Is `chargeTokenId` the same as `chargeId` or is the change from one to another intentional?
(given the `logSystemError` function it looks like it should have been `chargeId` from the start)

app/controllers/secure_controller.js
```diff
@@ -20,11 +20,14 @@ const paths = require('../paths')
 exports.new = async function (req, res) {
   const chargeTokenId = req.params.chargeTokenId || req.body.chargeTokenId
   const correlationId = req.headers[CORRELATION_HEADER] || ''
+  var chargeId
+  var gatewayAccountId
+  var gatewayAccountType
```
jankowiakmaria: [minor] `var` -> `let`

app/controllers/secure_controller.js
oswaldquek: `chargeTokenId` is different from `chargeId`. yup, it should have been `chargeId` from the start, which this PR fixes.

app/controllers/secure_controller.js
oswaldquek: what's the advantage of `let` here?

app/controllers/secure_controller.js
jankowiakmaria: That's what I though (but wanted to confirm). 👍

app/controllers/secure_controller.js
jankowiakmaria: In this particular case it doesn't make functional difference.
`let` is always more appropriate to use in terms of scoping. (`var` should really be avoided).
It's a minor comment - it doesn't block this PR. I was concerned about the other thing (that's why I haven't approved previously).

PR #1039, opened 2019-09-24
PP-5550: Minor simplification of charge_controller

oswaldquek:
Writing a unit test for PP-5550 is complicated. Possible - but complicated.
This is because the charge_controller's "create" method is unwieldy, too big
and overly complicated. Shockingly there are no unit tests for the method.
Unsurprising given the complexity of the code.

In order to write a unit test for PP-5550 I've decided to make small
incremental simplifications and eventually move the validation code into its
own class. But for now, this method gets rid of the 'let userEmail' as its
result is never used.

I have also taken a decision - in conjunction with @mrlumbu and @sfount - to
get rid of some AWSXray code. It makes code look more complicated; and in
reality we use request ids and charge ids to trace a request through the
system.

## WHAT

## HOW 
_Steps to test or reproduce:_

### DIFF
--- a/app/controllers/charge_controller.js
+++ b/app/controllers/charge_controller.js
@@ -112,7 +112,7 @@ module.exports = {
         responseRouter.response(req, res, 'SYSTEM_ERROR', withAnalytics(charge))
       })
   },
-  create: (req, res) => {
+  create: async (req, res) => {
     const charge = normalise.charge(req.chargeData, req.chargeId)
     const cardModel = Card(req.chargeData.gateway_account.card_types, req.headers[CORRELATION_HEADER])
     const chargeOptions = {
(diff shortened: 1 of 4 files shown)


oswaldquek:
@sfount @heathd @alexbishop1 @mrlumbu I'd appreciate if two or more of you can have a look at this

app/controllers/charge_controller.js
```diff
@@ -129,81 +127,72 @@ module.exports = {
...
-              emailChanged = req.body.originalemail !== userEmail
-            }
-            emailTypos = commonTypos(userEmail)
-            if (req.body['email-typo-sugestion']) {
-              userEmail = emailChanged ? req.body.email : req.body['email-typo-sugestion']
-              emailTypos = req.body['email-typo-sugestion'] !== req.body.originalemail ? commonTypos(userEmail) : null
-            }
-            emailPatch = Charge(req.headers[CORRELATION_HEADER]).patch(req.chargeId, 'replace', 'email', userEmail, subSegment)
+        if (charge.gatewayAccount.emailCollectionMode !== 'OFF' ||
+          (charge.gatewayAccount.emailCollectionMode !== 'OPTIONAL' && req.body.email && req.body.email !== '')
```
stephencdaly: An empty string is a falsy value, so we can get rid of `req.body.email !== ''` I think

app/controllers/charge_controller.js
```diff
@@ -129,81 +127,72 @@ module.exports = {
...
-          redirect(res).toNew(req.chargeId)
-        })
-        .then(data => {
-          subSegment.close()
-          card = data.card
+    validator.verify(req)
+      .catch(() => {
+        redirect(res).toNew(req.chargeId)
+      })
+      .then(async data => {
```
stephencdaly: I'm not sure I like mixing the `.then(...)` and async/await syntax in the same function and maybe in the same class. But if we're incrementally improving, then maybe doing it for the entire enclosing function would be the best way to go.

app/controllers/charge_controller.js
stephencdaly: Also not sure this conditional is quite right, it will always return true unless collection more is 'OFF'.
Maybe it should check if collection mode is 'ON' or 'OPTIONAL' and email is provided.

app/controllers/charge_controller.js
oswaldquek: I agree. I tried to make `validator.verify(req)` async but I wasn't quite sure how. maybe you could help me out with that.

app/controllers/charge_controller.js
oswaldquek: fixed

app/controllers/charge_controller.js
```diff
@@ -129,81 +127,70 @@ module.exports = {
...
-        .then(data => {
-          subSegment.close()
-          card = data.card
+    let data
+    try {
+      data = await validator.verify(req)
+    } catch (err) {
+      return redirect(res).toNew(req.chargeId)
+    }
+    card = data.card
```
stephencdaly: can declare this as a `const` here rather than using `let` earlier up now

app/models/charge.js
```diff
@@ -89,26 +89,22 @@ module.exports = correlationId => {
...
-        })
-    })
+      const response = await connectorClient({ correlationId }).patch({ chargeId, payload }, subSegment)
+      if (response.statusCode === 200) {
+        return
+      } else {
+        throw new Error('Calling connector to patch a charge returned an unexpected status code')
+      }
+    } catch (err) {
+      throw err
```
stephencdaly: does catching and then throwing do something different to doing nothing?

app/models/charge.js
```diff
@@ -89,26 +89,22 @@ module.exports = correlationId => {
...
-        })
-        .catch(err => {
-          reject(err)
-        })
-    })
+      const response = await connectorClient({ correlationId }).patch({ chargeId, payload }, subSegment)
+      if (response.statusCode === 200) {
+        return
+      } else {
+        throw new Error('Calling connector to patch a charge returned an unexpected status code')
```
stephencdaly: could inverse the if statement so we just throw an error if the status code isn't 200, the we can do without the return statement

app/models/charge.js
oswaldquek: i'm not sure. the code used to be
```.catch(err => {	
          reject(err)	
        })
```
so I assumed throwing the `err` was enough. I'll throw a new Error as it seems that's the pattern though.

app/models/charge.js
oswaldquek: fixed

app/models/charge.js
```diff
@@ -89,26 +89,20 @@ module.exports = correlationId => {
...
-        .catch(err => {
-          reject(err)
-        })
-    })
+      const response = await connectorClient({ correlationId }).patch({ chargeId, payload }, subSegment)
+      if (response.statusCode !== 200) {
+        throw new Error('Calling connector to patch a charge returned an unexpected status code')
+      }
+    } catch (err) {
+      throw new Error('CLIENT_UNAVAILABLE', err)
```
sfount: Is this behaviour not different now by passing `CLIENT_UNAVAILABLE`? Without good unit tests this might cause an issue with upstream handling, no idea though.

app/models/charge.js
oswaldquek: @sfount As the old code did a `reject(new Error('Calling connector to patch a charge returned an unexpected status code'))` I don't think anything's been changed here.

app/models/charge.js
oswaldquek: the old code did
```
        .catch(err => {	
          reject(err)	
        })
```
would it better to just do `throw err` instead of `throw new Error('CLIENT_UNAVAILABLE', err)` here?
It's going to get caught in charge_controller.js like so:
```
      try {
        await Charge(req.headers[CORRELATION_HEADER]).patch(req.chargeId, 'replace', 'email', userEmail)
      } catch (err) {
        return responseRouter.response(req, res, 'SYSTEM_ERROR', withAnalytics(charge))
      }
```

sfount: I think keeping the same behaviour would be good for something that's solely a refactor. 

My understanding is that if you remove the try/ catch entirely it will behave the same (as `err => { reject(err) }`)

oswaldquek:
@sfount there weren't any unit tests for charge_controller's `create` method to begin with, so i've relied on the cypress tests and e2e tests passing for confidence here. I have a unit test for this but it will be in the next PR to keep PRs small.

app/models/charge.js
stephencdaly: I reckon it had to do this because the code was manually returning a promise, but now we're using async/await which automatically wraps the function in a promise so any errors thrown will cause the promise to be rejected.

app/models/charge.js
oswaldquek: @sfount ah, I didn't know that. I've removed the try catch now.

app/models/charge.js
```diff
@@ -89,26 +89,16 @@ module.exports = correlationId => {
...
-            resolve()
-          } else {
-            reject(new Error('Calling connector to patch a charge returned an unexpected status code'))
-          }
-        })
-        .catch(err => {
-          reject(err)
-        })
-    })
+  const patch = async function (chargeId, op, path, value, subSegment) {
```
sfount: 👍

app/controllers/charge_controller.js
```diff
@@ -112,98 +112,84 @@ module.exports = {
...
-    }, clsSegment)
+        data.validation.typos = emailTypos
+        data.validation.originalEmail = userEmail
+      }
+      charge.countries = countries
+      try {
+        await appendChargeForNewView(charge, req, charge.id)
+      } catch (err) {
+        logging.failedGetWorldpayDdcJwt(err)
+        return responseRouter.response(req, res, 'SYSTEM_ERROR', withAnalytics(charge))
```
sfount: I'm uncomfortable with the number of places that this method can terminate, there are ~5 and it makes it almost impossible to know how many combinations of things can happen. 

I know this isn't part of this PR but a much easier to reason about structure here would be to run validation, come to a conclusion and then only respond once - either with success or failure. Responding in-line makes it hard to know if this will do exactly what the previous code did.

oswaldquek: i agree... i might do more refactoring to reduce the termination places but the 
```
        logging.failedGetWorldpayDdcJwt(err)
        return responseRouter.response(req, res, 'SYSTEM_ERROR', withAnalytics(charge)
```
was there before.

PR #1047, opened 2019-09-26
Pp 5550 test for worldpay3ds flex ddc result

oswaldquek:
Send worldpay_3ds_flex_ddc_result to connector depending on whether it is present in the body of the POST to `/card_details/PAYMENT_ID` on frontend.


test/controllers/charge_controller_create_test.js
```diff
@@ -0,0 +1,199 @@
...
+const chargeId = '42mdrsshtsk4chpeoifhlgf4lk'
+
+const card = {
+  'brand': 'visa',
+  'type': 'CREDIT',
+  'corporate': false,
+  'prepaid': 'NOT_PREPAID'
+}
+
+const chargeData = {
```
stephencdaly: I think we should use `payment_fixtures`#`validChargeDetails().getPlain()` to construct this JSON. Having the connector response structure in one place makes the test code easier to maintain and the fixtures are also used by pact tests so we have some confidence that the JSON structure they produce is valid.

test/controllers/charge_controller_create_test.js
```diff
@@ -0,0 +1,199 @@
...
+const path = require('path')
+
+// NPM dependencies
+const proxyquire = require('proxyquire')
+const sinon = require('sinon')
+const expect = require('chai').expect
+
+const chargeId = '42mdrsshtsk4chpeoifhlgf4lk'
+
+const card = {
```
stephencdaly: As with the comment below, we should use `payment_fixtures`#`validCardDetails` to construct this JSON

test/controllers/charge_controller_create_test.js
```diff
@@ -0,0 +1,199 @@
...
+          'corporate_card': card.corporate,
+          'prepaid': card.prepaid,
+          'address': {
+            'line1': paymentDetails.addressLine1,
+            'city': paymentDetails.addressCity,
+            'postcode': paymentDetails.addressPostcode,
+            'country': paymentDetails.addressCountry
+          },
+          'worldpay_3ds_flex_ddc_result': paymentDetails.worldpay3dsFlexDdcResult
+        }
```
stephencdaly: I think we should use `payment_fixtures`#`validAuthorisationRequest` to construct this JSON.

oswaldquek: I've done this but it's increased the number of lines of code as I've had to delete the accept_header and user_agent_header from the construct returned by `validAuthorisationRequest`. This is necessary as I've mocked out the calls to `req.header` with `header: sinon.spy()` earlier on in the test.

PR #1092, opened 2019-10-10
WIP - if production then get all static assets from CDN

jonheslop:
If I were a developer on GOV.UK Pay I would simply setup a pipeline to port all assets over to a CDN, hash the file names and merge this PR.


sfount:
Static asses now cached by the Nginx proxy which is much better at serving static files, as this is in front of the server no URL changes needed.

This was mostly on the nginx proxy instance repository but https://github.com/alphagov/pay-frontend/pull/1972 includes changes to help this work correctly.

Fully moving over to CDN would further reduce the load on our provisioned services, this may be proposed in future work.

PR #1108, opened 2019-10-21
PP-5715 Switch to structured logging

stephencdaly:
Bump Winston to v3

Log in JSON format, including the standard keys included for all log lines

Also up the node version to v8.12.0 so that Husky will allow me to commit. Was getting the error:
lint-staged requires at least version 8.12.0 of Node, please upgrade


stephencdaly:
Snyk is complaining about a dependency of Nunjucks:
```
Vulnerable module: mixin-deep
Introduced through: nunjucks@3.2.0
```
And says the remediation is upgrading to v3.2.0, which is what we're on 🤔 
This is only a dev dependency though... @jonheslop any thoughts?

stephencdaly:
Hmm me running `npm install` has downgraded mixin-deep from 1.3.2 to 1.3.1, any idea why it would do that @jonheslop ?

stephencdaly:
@danworth Snyk is now happy

stephencdaly:
Ah but now the build is failing :( stupid npm

PR #1143, opened 2019-11-04
PP-5778 Log part of paRes

kbottla:
## WHAT
- paRes received during authorisation can be upto 4kb, but docker truncates long log lines as it doesn't support more than 2kb.
- Only log the beginning and last of paRes so we can check if connector receives full paRes for failed payments

Related to https://github.com/alphagov/pay-frontend/pull/1141

### DIFF
--- a/app/controllers/three_d_secure_controller.js
+++ b/app/controllers/three_d_secure_controller.js
@@ -31,7 +31,7 @@ const build3dsPayload = (chargeId, req) => {
   const paRes = _.get(req, 'body.PaRes')
   if (!_.isUndefined(paRes)) {
     auth3dsPayload.pa_response = paRes
-    logger.info(`paRes received for charge [${chargeId}] 3DS authorisation [${paRes}]`)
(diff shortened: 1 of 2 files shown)


app/controllers/three_d_secure_controller.js
```diff
@@ -31,7 +31,7 @@ const build3dsPayload = (chargeId, req) => {
   const paRes = _.get(req, 'body.PaRes')
   if (!_.isUndefined(paRes)) {
     auth3dsPayload.pa_response = paRes
-    logger.info(`paRes received for charge [${chargeId}] 3DS authorisation [${paRes}]`)
+      logger.info(`paRes received for charge [${chargeId}] 3DS authorisation [starts with [${paRes.substring(0,50)}] and ending with [${paRes.substring(paRes.length-50)}] ]`)
```
jankowiakmaria: What if length is < 50?
(when argument passed to substring is < 0 it will be treated as 0 - is it desired behaviour?)

jankowiakmaria: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring

kbottla: yes, for the log lines we are interested in paRes with length > 4000 which are failing 3DS payments

PR #1151, opened 2019-11-06
PP-5709: add ip address to auth request

jankowiakmaria:
It's required for connector to add appropriate attribute to
Worldpay authorisation request. Worldpay can then use it for
fraud checks.

Changes:
* add user_ip_address helper (code taken from https://github.com/alphagov/pay-selfservice/blob/dc4be64faab6b2b141bb54df1d216d971256496f/app/controllers/request-to-go-live/agreement/post.controller.js#L25-L36) plus relevant unit test
* update normalise_charge to include ip_address property when creating connector
request
* update tests to reflect changes (include ip_address)
* rename file name with a typo


app/services/normalise_charge.js
```diff
@@ -145,20 +146,21 @@ module.exports = (function () {
...
-      'cvc': req.body.cvc,
-      'card_brand': card.brand,
-      'expiry_date': expiryDate(req.body.expiryMonth, req.body.expiryYear),
-      'cardholder_name': req.body.cardholderName,
-      'card_type': card.type,
-      'corporate_card': card.corporate,
-      'prepaid': card.prepaid,
-      'address': addressForApi(req.body),
-      'accept_header': req.header('accept'),
-      'user_agent_header': req.header('user-agent')
```
jankowiakmaria: it seems my editor removed all `'`.
Let me know if it's an issue and I should roll it back.

jankowiakmaria:
Pact test failing - unrecognised property; I guess I need to update connector first. (it doesn't prevent from review though)

app/services/normalise_charge.js
sfount: I think this would be good to do ^ 

I approve of removing the `'` but it should be done as a separate PR so that it's clear what has actually changed for this work.

jankowiakmaria: Sure - will clean it up.

jankowiakmaria: Nope, it's not my editor, it's pre-commit hook (linter) that "fixes" the files. Should I fight it?

sfount: ⚔️ 👍

jankowiakmaria:
Ready to be reviewed

app/utils/user_ip_address.js
```diff
@@ -0,0 +1,14 @@
+'use strict'
+
+module.exports = req => {
+  const xHeaderIpAddress = req.headers['x-forwarded-for']
+  let ipAddress
+  if (xHeaderIpAddress !== undefined) {
+    ipAddress = xHeaderIpAddress.split(',')[0]
+  } else {
+    ipAddress = (req.connection && req.connection.remoteAddress) ||
```
RoryMalcolm: idk maybe a little comment here to say what this logic does?

app/services/normalise_charge.js
```diff
@@ -155,7 +156,8 @@ module.exports = (function () {
       'prepaid': card.prepaid,
       'address': addressForApi(req.body),
       'accept_header': req.header('accept'),
-      'user_agent_header': req.header('user-agent')
+      'user_agent_header': req.header('user-agent'),
+      'ip_address': userIpAddress(req)
```
sfount: I know this may be copied over from another service but I wonder if we could name `userIpAddress` more meaningfully. Especially as it's in the generic utils folder. 

It seems like it's used to extract the source ip address from the req object no matter where it is.

jankowiakmaria: `source_ip_address`? Suggestions?

app/utils/user_ip_address.js
jankowiakmaria: comments are risky (hard to maintain, code _should_ be self-explanatory) - and I don't think it's required in this case.

app/services/normalise_charge.js
sfount: I don't have any suggestions here I just don't know what calling the function `userIpAddress` might do, especially as it's imported directly as a module. 

Some patters might look like 
`utils/IPParser` -> `IPParser.getIpFromRequest(req)` 

I didn't request changes on this review as it's not dissimilar to other code, I just think we need to think about our naming more carefully on our Node apps.

jankowiakmaria: Sure - I don't mind renaming. I just expect suggestions when I'm asked to do so :) (that's it)

PR #1162, opened 2019-11-13
BAU Include parent layout start block in payment page

sfount:
Cookie banner currently isn't shown on the card payment page. 
* include parent layout block to show this

We should verify that the cookie banner should be shown on payment pages, however this brings it in line with all of the other payment journey (3ds, confirmation, etc.)


sfount:
I think this is the most correct thing to do from a code structure view but questions around cookie banner behaviour on payment pages should be answered and reviewed through a ticket.

PR #1171, opened 2019-11-14
BAU Add (TMP) CSPs for card payment page

sfount:
* adds helmet 
* basic configuration for reporting on CSP violation (report only) 
* new keys have been generated for this test waiting on the decision to use sentry or not


sfount:
That was cleaned up and properly implemented in https://github.com/alphagov/pay-frontend/pull/1172

PR #1184, opened 2019-11-21
PP-5851 Log when session variable deleted from cookie

stephencdaly:


mrlumbu:
It would be good to say why you're doing this in commit, so future us can reason about whether it's still needed.

PR #1195, opened 2019-11-26
PP-5872 Browserify Card details page inline scripts

kbottla:
## WHAT
- Externalise `address country` autocomplete & setting global charge ID scripts to helper.js, so we don't need to allow unsafe-line scripts as part of CSP
- Add exception for `openregisterLocationPicker` as linter complains it as 'not defined'. `openregisterLocationPicker` is defined in a script (`location-autocomplete.min.js`)  loaded separately

There is no visible impact (when running locally) on performance due to externalisation of these scripts

**Before the change**
Before - desktop

**and After the change**
After - desktop

### DIFF
--- a/app/views/charge.njk
+++ b/app/views/charge.njk
@@ -492,22 +492,9 @@
     </aside>
   </div>
 </div>
-  <script id="govuk-script-charge">
-    window.chargeId = "{{ id }}"
-  </script>
+
 {% endblock %}
 
 {% block bodyEnd %}
-  {% if service.collectBillingAddress %}
-  <script type="text/javascript" src="/public/location-autocomplete.min.js"></script>
-  <script type="text/javascript">
-    openregisterLocationPicker({
-      selectElement: document.getElementById('address-country'),
-      url: '/public/countries-autocomplete-graph.json',
-      autoselect: true,
-      displayMenu: 'overlay'
-    })
-  </script>
-  {% endif %}
   {% include "includes/scripts.njk" %}
 {% endblock %}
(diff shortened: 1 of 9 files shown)


app/views/charge.njk
```diff
@@ -492,22 +492,11 @@
...
-  <script type="text/javascript">
-    openregisterLocationPicker({
-      selectElement: document.getElementById('address-country'),
-      url: '/public/countries-autocomplete-graph.json',
-      autoselect: true,
-      displayMenu: 'overlay'
-    })
-  </script>
-  {% endif %}
+  <input id="collect-billing-address" name="collectBillingAddress" type="hidden" value="{{ service.collectBillingAddress }}"/>
```
sfount: As discussed we might be able to keep very similar behaviour by separating and templating in the initialise method into `scripts.njk` -- this would mean there are no different behaviours introduced. 

It would look something like: 
```js
{% if service.collectBillingAddress %}
window.payScripts.helpers.loadLocationAutocompleteScript()
{% endif %}
```

kbottla: 👍 Updated

app/views/includes/scripts.njk
```diff
@@ -40,6 +40,10 @@
     window.GOVUKFrontend.initAll();
     if (mainWrap.classList.contains('charge-new')) {
       showCardType().init();
+      window.payScripts.helpers.setGlobalChargeId();
```
sfount: As discussed it looks like the `showCardType()` scrits end up using this global chargeId - it would be good for this to be as high up as possible to make sure everything has access.

kbottla: Updated..

PR #1197, opened 2019-11-26
PP-5873 turn CSP on with report only option

SandorArpa:
## WHAT
- Turn CSP protection on with report only option
- add required exceptions
- 'unsafe-inline' from scriptSrc shall be removed once [PP-5872](https://payments-platform.atlassian.net/browse/PP-5872) completed

with @sfount

### DIFF
--- a/app/middleware/csp.js
+++ b/app/middleware/csp.js
@@ -7,9 +7,45 @@ const environment = process.env.ENVIRONMENT
 
 const sentryCspReportUri = `${cspReportUri}&sentry_environment=${environment}`
 
+// Worldpay 3ds flex iframe
+const frameSource = ["'self'", 'https://secure-test.worldpay.com/']
+
+// Google analytics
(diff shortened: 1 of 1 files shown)


app/middleware/csp.js
```diff
@@ -7,9 +7,45 @@ const environment = process.env.ENVIRONMENT
...
 
+// Worldpay 3ds flex iframe
+const frameSource = ["'self'", 'https://secure-test.worldpay.com/']
+
+// Google analytics
+const imgSource = ["'self'", 'https://www.google-analytics.com/', 'http://www.google-analytics.com/']
+
+// Google analytics
+const scriptSource = ["'self'", 'https://www.google-analytics.com/', 'http://www.google-analytics.com/',
+  "'unsafe-inline'"]
```
sfount: `unsafe-inline` is temporary pending externalising frontend scripts. This will let us test all external rules and then restrict scripts and apply non-report only once they have been moved out.

PR #1198, opened 2019-11-26
PP-5831 CSP exceptions for non-external scripts

sfount:
A number of the inline scripts on the card payment page are dynamically
templated in based on various params that can change per journey. This
makes them complex to safely hash.

Use a `nonce` generated per request to initially allow rejected
`unsafe-inline` policy.

---

As we extend the base design system `layout` template we inherit a
script to set a 'js-enabled' class on themain page.

We don't have control to add a nonce here and the script never changes
so allow this particular hash.

### DIFF
--- a/app/middleware/csp.js
+++ b/app/middleware/csp.js
@@ -1,11 +1,15 @@
 const helmet = require('helmet')
 
-const sendCspHeader = true
-const enforceCsp = true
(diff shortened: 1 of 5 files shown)


sfount:
Closed in favour of "draft" proposal https://github.com/alphagov/pay-frontend/pull/1199 @alexbishop1

PR #1208, opened 2019-11-29
[DRAFT] PP-5885 Test CSP violations with cypress

kbottla:
- upgrades cypress
- Sets CSP header and enforces CSP for cypress tests
- checks if CSP violations are applied (set incorrect hash)


kbottla:
CSP doesn't work with Cypress

PR #1221, opened 2019-12-04
PP-5884 add csp restrictions to worldpay flex ddc initial iframe

SandorArpa:
## WHAT
- add csp headers to worldpay 3ds flex ddc frame
- refactor csp.js to have separate methods for different csp settings

### DIFF
--- a/app/middleware/csp.js
+++ b/app/middleware/csp.js
@@ -17,20 +17,21 @@ const CSP_NONE = ["'none'"]
 const CSP_SELF = ["'self'"]
 
 // Worldpay 3ds flex iframe - frame and child must be kept in sync
-const frameAndChildSource = ["'self'", 'https://secure-test.worldpay.com/', 'https://centinelapi.cardinalcommerce.com/']
+const frameAndChildSourceCardDetails = ["'self'", 'https://secure-test.worldpay.com/', 'https://centinelapi.cardinalcommerce.com/']
 
-const imgSource = ["'self'", 'https://www.google-analytics.com/']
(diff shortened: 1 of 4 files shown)


app/middleware/csp.js
```diff
@@ -17,20 +17,21 @@ const CSP_NONE = ["'none'"]
...
+const frameAndChildSourceCardDetails = ["'self'", 'https://secure-test.worldpay.com/', 'https://centinelapi.cardinalcommerce.com/']
 
-const imgSource = ["'self'", 'https://www.google-analytics.com/']
+const imgSourceCardDetails = ["'self'", 'https://www.google-analytics.com/']
 
-const scriptSource = ["'self'", 'https://www.google-analytics.com/',
+const scriptSourceCardDetails = ["'self'", 'https://www.google-analytics.com/',
   (req, res) => `'nonce-${res.locals && res.locals.nonce}'`, govUkFrontendLayoutJsEnabledScriptHash]
 
+const formActionWP3DS = ["'self'", 'https://centinelapi.cardinalcommerce.com/V1/Cruise/Collect', 'https://secure-test.worldpay.com/shopper/3ds/ddc.html']
```
sfount: Should this just be the domain that we trust? 
`https://secure-test.worldpay.com/shopper/3ds/ddc.html`

It seems strange that this is tied to a `.html` route, maybe just `shopper/3ds` will be sufficient for this rule being secure.

SandorArpa: 👍

PR #1228, opened 2019-12-05
PP-5902 add gstatic site to csp img-src directive

SandorArpa:
## WHAT
- https://www.gstatic.com/instantbuy/icons/ serves an image that required by GPay and thus we need to whitelist it.

### DIFF
--- a/app/middleware/csp.js
+++ b/app/middleware/csp.js
@@ -20,7 +20,7 @@ const CSP_SELF = ["'self'"]
 const frameAndChildSourceCardDetails = ["'self'", 'https://secure-test.worldpay.com/',
   'https://centinelapi.cardinalcommerce.com/']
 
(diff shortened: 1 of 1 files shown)


app/middleware/csp.js
```diff
@@ -20,7 +20,8 @@ const CSP_SELF = ["'self'"]
 const frameAndChildSourceCardDetails = ["'self'", 'https://secure-test.worldpay.com/',
   'https://centinelapi.cardinalcommerce.com/']
 
-const imgSourceCardDetails = ["'self'", 'https://www.google-analytics.com/']
+const imgSourceCardDetails = ["'self'", 'https://www.google-analytics.com/',
+  'https://www.gstatic.com/instantbuy/icons/']
```
alexbishop1: I think we can probably just whitelist `https://www.gstatic.com/` — paths were introduced in CSP2 so we might block people using older browsers. I doubt there’s a huge security difference.

PR #1236, opened 2019-12-06
PP-5930 prefetch-src directive for ZAP tests

kbottla:
## WHAT
- `prefetch-src` directive is an experimental and CSP level 3 directive which is not required on card details page.
       However, ZAP tests fails on directive being not defined.

- Adds a flag `CSP_ADD_PREFETCH_DIRECTIVE` to add prefetch-src directive to card details page (set to `true` to add directive and end to end environment only)


kbottla:
Closing as we don't want to add `prefetch-src` directive just for ZAP tests

PR #1243, opened 2019-12-09
PP-5762 Remove correlation id from log message

stephencdaly:
The correlation id is now added as a structured argument. Remove it from log messages where it is in square brackets at the start of the message for consistency.


oswaldquek:
sorry, could you point me out to where the correlation id is set in structured logging now?

stephencdaly:
> sorry, could you point me out to where the correlation id is set in structured logging now?

in the `correlation_header` middleware, we add it to the logging fields on the request which are then added into the logs in all the places we do logging.

PR #1246, opened 2019-12-10
PP-5762 Add appropriate keys to log lines

stephencdaly:
Add keys to log lines containing useful information, such as the error.

Make it so that when we log a http status code, we always use "status_code" as the key.


stephencdaly:
> Looks good, could `status_code` be a common key added to js/java-commons?

Yeh maybe. But in Javascript it's actually quite annoying using constants for keys added ad-hoc in log lines as there's not a very syntactically nice way to do it. So I'm not entirely sure the benefit is worthwhile.

PR #1247, opened 2019-12-10
PP-5933 add form action covering direct redirect feature

SandorArpa:
## WHAT
- redirect is blocked on some browsers when the site that we're redirecting to is redirecting to a different domain, eg on cancel we redirect the user to example.com/abc and that site redirects the user to example.org/abc. hence the need to use * for form action in case the redirectToServiceImmediatelyOnTerminalState flag is true
- if redirectToServiceImmediatelyOnTerminalState is false, we let the to be posted to self and products (when the user is coming from a payment link, the page automatically redirects the user back to products)

### DIFF
--- a/app/middleware/csp.js
+++ b/app/middleware/csp.js
@@ -12,7 +12,6 @@ const sentryCspReportUri = `${cspReportUri}&sentry_environment=${environment}`
 // and never changes
 const govUkFrontendLayoutJsEnabledScriptHash = "'sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU='"
 
-// Worldpay 3ds flex iframe
 const CSP_NONE = ["'none'"]
 const CSP_SELF = ["'self'"]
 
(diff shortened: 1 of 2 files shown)


app/middleware/csp.js
```diff
@@ -27,6 +27,16 @@ const scriptSourceCardDetails = ["'self'", "'unsafe-inline'", 'https://www.googl
...
 const formActionWP3DS = ["'self'", 'https://centinelapi.cardinalcommerce.com/V1/Cruise/Collect',
   'https://secure-test.worldpay.com/shopper/3ds/ddc.html']
+
+// Direct redirect use case lets post to any given site
+const formActionCardDetails = (req, res) => {
+  if (res.locals && res.locals.service &&
+      res.locals.service.redirectToServiceImmediatelyOnTerminalState === true) {
+    return '*'
+  }
+  return CSP_SELF + ', \'https://products.payments.service.gov.uk\''
```
sfount: This could avoid `\` escaping and `+` with string interpolation: 
```js
return `${CSP_SELF}, 'https://products.payments.service.gov.uk'`
```

PR #1254, opened 2019-12-13
PP-5940 Hide Wallet/Error containers during Worldpay 3DS flex payments

kbottla:
## WHAT
- When submitting card details to Worldpay flex, wallet options (google/apple pay) & any previous error messages were still shown
  to users.
- Changes hides wallet options and error container if Worldpay 3DS flex payment is in progress

with @sandorarpa

### DIFF
--- a/app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
+++ b/app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
@@ -3,6 +3,18 @@
 const toggleWaiting = () => {
   document.getElementById('card-details-wrap').classList.toggle('hidden')
   document.getElementById('spinner').classList.toggle('hidden')
+  document.getElementById('error-summary').classList.add('hidden')
+
+  var paymentMethodSubmitElement = document.getElementById('payment-method-submit')
+  if (typeof paymentMethodSubmitElement !== 'undefined' && paymentMethodSubmitElement !== null) {
+    paymentMethodSubmitElement.classList.add('hidden')
+  }
+
+  var payDividerElements = document.getElementsByClassName('pay-divider')
+  if (typeof payDividerElements !== 'undefined' && payDividerElements.length > 0) {
+    payDividerElements[0].classList.remove('pay-divider')
+    payDividerElements[0].classList.add('hidden')
+  }
 }
 
 const addWorldpaySessionIdToForm = (form, worldpaySessionId) => {


app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
```diff
@@ -3,6 +3,23 @@
...
+
+  var payDividerElements = document.getElementsByClassName('pay-divider')
+  if (typeof payDividerElements !== 'undefined' && payDividerElements.length > 0) {
+    payDividerElements[0].classList.remove('pay-divider')
+    payDividerElements[0].classList.toggle('hidden')
+  }
+
+  var errorSummaryContainer = document.getElementById('error-summary')
+  if (typeof errorSummaryContainer !== 'undefined' && errorSummaryContainer !== null &&
+    !errorSummaryContainer.classList.contains('hidden')) {
```
sfount: As we can't guarantee the current `hidden` state of these elements, should we just use the `add` method in favour of `toggle`? That way we can't accidentally 'unhide' something that was supposed to stay hidden. 

It would mean we wouldn't have to check for the existence of hidden and make this something like: 
```js
document.getElementById('error-summary').classList.add('hidden')
```

kbottla: 👍 Updated https://github.com/alphagov/pay-frontend/commit/d943f2e416bf5ebac36e5704e328d19bab257854

app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
```diff
@@ -3,6 +3,18 @@
...
+  document.getElementById('error-summary').classList.add('hidden')
+
+  var paymentMethodSubmitElement = document.getElementById('payment-method-submit')
+  if (typeof paymentMethodSubmitElement !== 'undefined' && paymentMethodSubmitElement !== null) {
+    paymentMethodSubmitElement.classList.add('hidden')
+  }
+
+  var payDividerElements = document.getElementsByClassName('pay-divider')
+  if (typeof payDividerElements !== 'undefined' && payDividerElements.length > 0) {
+    payDividerElements[0].classList.remove('pay-divider')
```
sfount: Concerned that this will behave differently if any other dividers are added to the page. This could use an identifier to make it not ambiguous and future proof.

kbottla: yes possible. Will look into separately.

PR #1266, opened 2019-12-18
zd3873417: Update pay-js-commons

oswaldquek:
## WHAT

## HOW 
_Steps to test or reproduce:_

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     ]
   },
   "dependencies": {
-    "@govuk-pay/pay-js-commons": "2.19.1",
+    "@govuk-pay/pay-js-commons": "2.20.0",
     "@sentry/node": "5.10.2",
     "appmetrics": "5.1.1",
     "appmetrics-statsd": "3.0.0",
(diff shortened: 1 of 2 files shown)


stephencdaly:
probably need to commit package-lock.json after running `npm install`

package.json
```diff
@@ -122,7 +122,7 @@
     "chai-string": "^1.4.0",
     "cheerio": "^1.0.0-rc.3",
     "chokidar-cli": "latest",
-    "cypress": "3.1.5",
+    "cypress": "^3.8.0",
```
stephencdaly: I think we prefer to keep dependencies anchored to a specific version nowadays as dependabot will upgrade versions. So remove the `^` I think

oswaldquek: This happened automatically as a result of running `npm audit fix` but sure I'll remove it.

stephencdaly: looks like browser tests have failed, so we probably can't upgrade

stephencdaly:
I think we use a docker image of cypress on CI and the version probably has to match, so don't think upgrading cypress works without some more effort.

PR #1267, opened 2019-12-18
PP-5977 Remove Google Analytics Front end

nimalank7:
Description:
- Complying with ICO guidance around cookie consent we can't set non-essential cookies without consent being granted. In the short term then we have to disable Google Analytics as it stores cookies

- Uncomment when we have enabled opt-in and remove the empty ga() function
- Renable analytics_ui_tests.js when Google Analytics is enabled


nimalank7:
JS comments removed and replaced with nunjuncks comments

PR #1325, opened 2020-01-16
PP-6007: Awaiting auth test

oswaldquek:
## WHAT

## HOW 
_Steps to test or reproduce:_

### DIFF
--- a/test/cypress/integration/card/awaiting-auth.spec.js
+++ b/test/cypress/integration/card/awaiting-auth.spec.js
@@ -1,0 +1,113 @@
+const cardPaymentStubs = require('../../utils/card-payment-stubs')
+
+const tokenId = 'be88a908-3b99-4254-9807-c855d53f6b2b'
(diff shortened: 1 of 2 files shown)


test/cypress/integration/card/awaiting-auth.spec.js
```diff
@@ -0,0 +1,106 @@
...
+      addressLine1: validPayment.addressLine1,
+      city: validPayment.city,
+      postcode: validPayment.postcode,
+      email: validPayment.email
+    }
+
+    const confirmPaymentDetailsStubs = [
+      { name: 'adminUsersGetService', opts: {} },
+      {
+        name: 'connectorMultipleSubsequentChargeDetails',
```
stephencdaly: It might be good to add a comment to reason why we need this stub to return the different charge statuses on subsequent requests. I can see it causing some headscratching in the future.

PR #1352, opened 2020-01-30
PP-6031 Bump engine version

danworth:
PaaS node buildpack no longer supports 12.11.1. To avoid having to keep
bumping the engine version whilst maintaining compatibility with our
existing docker image build process, this pins the major version whilst
accepting minor and pataches releases greater than 12.11.1.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "private": true,
   "license": "MIT",
   "engines": {
-    "node": "12.11.1"
+    "node": "^12.11.1"
   },
   "standard": {
     "globals": [


richardTowers:
Will this change the node version in the current infrastructure too? Presumably that's a docker thing, so probably "no"?

If it will, do we need to do any additional testing that this isn't going to break anything? It's just a minor version bump, and node's pretty stable, so it's probably not going to break I guess.

danworth:
After discussion we've decided to pin the major version and allow a range on the minor and patch. Node buildpack should pick the highest 12.x release and our docker build process should be satisfied that the existing 12.11.1 version of node in our base image is compliant with the range.

danworth:
Just to be sure. I've built the image locally with this change, run a container with it and checked the node version:
```
gds5062 > docker build . --tag my_frontend
gds5062 > docker run -it --entrypoint=/bin/sh my_frontend -i
/app # node --version
v12.2.0
```
The version of node matches that seen in the latest-master frontend:

```
gds5062 > docker run -it --entrypoint=/bin/sh govukpay/frontend:latest-master -i
/app # node --version
v12.2.0
```

Based on the above it seems this won't inadvertently bump the node version for our existing docker based deployments.

PR #1367, opened 2020-02-04
PP-6079 Don't collect billing address for moto payments

stephencdaly:
Currently, we don't show fields to collect the billing address on the card details page if billing address collection is disabled for the service.

Now, we also won't show these fields if a payment is a MOTO payment even if billing address collection is enabled for the service. Determine whether we should collect the address in the resolve_service middleware and add a field to res.locals so that we can access it in all templates
and controllers that need this value.

Add Cypress tests to test this and remove tests from charge_ui_tests which are now better covered by Cypress tests.

### DIFF
--- a/app/controllers/charge_controller.js
+++ b/app/controllers/charge_controller.js
@@ -122,7 +122,7 @@ module.exports = {
     const cardModel = Card(req.chargeData.gateway_account.card_types, req.chargeData.gateway_account.block_prepaid_cards, req.headers[CORRELATION_HEADER])
     const chargeOptions = {
       email_collection_mode: charge.gatewayAccount.emailCollectionMode,
-      collect_billing_address: res.locals.service.collectBillingAddress
+      collect_billing_address: res.locals.collectBillingAddress
     }
     const validator = chargeValidator(i18n.__('fieldErrors'), logger, cardModel, chargeOptions, getLoggingFields(req))
 
@@ -185,7 +185,7 @@ module.exports = {
 
     const correlationId = req.headers[CORRELATION_HEADER] || ''
     const payload = normalise.apiPayload(req, card)
-    if (res.locals.service.collectBillingAddress === false) {
+    if (res.locals.collectBillingAddress === false) {
       delete payload.address
     }
     try {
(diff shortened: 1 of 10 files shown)


app/controllers/charge_controller.js
```diff
@@ -185,7 +185,7 @@ module.exports = {
 
     const correlationId = req.headers[CORRELATION_HEADER] || ''
     const payload = normalise.apiPayload(req, card)
-    if (res.locals.service.collectBillingAddress === false) {
+    if (res.locals.collectBillingAddress === false) {
```
stephencdaly: This was the line I forgot to change before. Was causing validation in connector to fail. It works now (tested with pay local), and added a test in test/controllers/charge_controller_create_test.js

test/controllers/charge_controller_create_test.js
```diff
@@ -148,12 +145,51 @@ describe('POST /card_details/{chargeId} endpoint', function () {
...
+      }
+    ))).to.be.true // eslint-disable-line
+  })
+
+  it('should not include billing address in authorisation request when billing address collection disabled', async function () {
+    response.locals.collectBillingAddress = false
+
+    const request = {
+      chargeData: chargeData,
+      body: paymentDetails,
```
oswaldquek: the `paymentDetails` here contain some address data. should it not have any address data at all? since there won't be address data from the frontend.

test/controllers/charge_controller_create_test.js
```diff
@@ -148,12 +145,51 @@ describe('POST /card_details/{chargeId} endpoint', function () {
...
+      cardType: card.type,
+      corporateCard: card.corporate,
+      prepaid: card.prepaid,
+      noBillingAddress: true
+    }).getPlain()
+
+    delete payload.accept_header
+    delete payload.user_agent_header
+
+    console.log('PAYLOAD ' + JSON.stringify(payload))
```
oswaldquek: do we need this line?

stephencdaly: oh yeh good spot, I'll remove

test/controllers/charge_controller_create_test.js
stephencdaly: Yeh good point. I'll update.

PR #1382, opened 2020-02-17
PP-6147: Upgrade node to v12.4.1

richardTowers:
WHAT
----

This is the latest version of node supported by GOV.UK PaaS. We need to
be on at least v12.9.0 to use AWS CloudFront's Field Level Encryption.

The issue with `statsd` we encountered when we upgraded to node v12.13.1
(see [PP-5917](https://payments-platform.atlassian.net/browse/PP-5917))
looks like it's fixed in v12.4.1 (according to [the node release notes](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V12.md#2020-01-07-version-12141-erbium-lts-bethgriggs)).

**NOTE**: I'm not sure how to test that the statsd issue hasn't
regressed, so this will need a look by someone with more understanding
of the issue.

HOW
---

Use your knowledge / testing to confirm this isn't going to break
things in production.


sfount:
Confirming that `12.14.1` addressed the issues described in https://payments-platform.atlassian.net/browse/PP-5917

See https://github.com/nodejs/node/releases/tag/v12.14.1 and https://github.com/nodejs/node/pull/30196

richardTowers:
Okay, lets :shipit: then!

PR #1388, opened 2020-02-18
PP-6147: Decrypt field level encryption fields

richardTowers:
WHAT
----

Two things:

1) Adds some decryption middleware which, when enabled, will decrypt any card
   data in the request.
2) Amends server side validation so that, when enabled, card data will not be
   populated in the response

The commit messages have some more details.

HOW
---

The most important thing to check is that this doesn't break the existing
infrastructure before we merge it - the end-to-end tests should be more than
enough to cover that.

PP-6147 won't be done until we've tested this in the new PaaS environment,
which requires some additional changes in pay-omnibus. We could wait for that
testing to happen before merging this if we wanted.


richardTowers:
@mrlumbu mentioned that I got one of the test names backwards - I think I've fixed that in f0dbfeeb

richardTowers:
We've run the end-to-end tests against this PR, tested it manually in`test` (to check that the AJAX call works with `x-www-form-url-encoded`), and tested it manually in the paas environment with Field Level Encryption.

We're pretty confident the code works now, so as long as people are happy review wise this is good to go.

test/middleware/decrypt_card_data_test.js
```diff
@@ -0,0 +1,211 @@
+const path = require('path')
+const sinon = require('sinon')
+const proxyquire = require('proxyquire')
+const { expect } = require('chai')
+
+describe('decryptCardData middleware', function () {
+  let next
+  beforeEach(function () {
+    next = sinon.stub()
+    next.promise = new Promise(resolve => next.callsFake(resolve))
```
richardTowers: This doesn't handle promise rejections when `next` is called with an error, so some of the tests are probably not working correctly.

richardTowers: Fixed in 1e0047d9

app/middleware/decrypt_card_data.js
```diff
@@ -0,0 +1,65 @@
...
+        const result = await decryptNode.decrypt(keyring, Buffer.from(cipherText, 'base64'))
+        req.body[fieldName] = result.plaintext.toString('utf8')
+      } catch (rejection) {
+        logger.error('Failed to decrypt', { fieldName })
+        throw rejection
+      }
+    })).then(() => {
+      next()
+    }).catch((err) => {
+      next(err)
```
sfount: This is absolutely how one should handle an error when writing middleware in a stack like this, my understanding is that these apps have a very dubious understanding of what error handling is and therefore might not actually have anything at the end of the stack that picks this up nicely. 

In that case it would default to Expresses error handler which would send plaintext with hopefully the stack redacted in production mode. 

Having had a quick look it looks like other middleware handle errors inline and respond immediately through the response object. This is not a good pattern and makes a response super hard to reason about but this PR might need to factor in that it would have to do that or add something at the end that will respond with errors in a uniform way.

sfount: My understanding of what you'd be expecting at the other end but not actually find in frontend https://github.com/alphagov/pay-toolbox/blob/master/src/web/errorHandler.ts#L64-L87

richardTowers: You're correct. Here's a screenshot with it snowflaked to error 😞 

image

richardTowers: This should be addressed in 8e1a1da - I've added a default error handler. I've tested this manually in the paas staging environment and now you get the nice:

```
We’re experiencing technical problems

No money has been taken from your account.
```

thing instead.

PR #1433, opened 2020-03-16
Read node version from package engines into nvmrc

rjbaker:


gidsg:
closing as we don't use travis anymore

PR #1438, opened 2020-03-17
PP-6259 Make frontend send user to do 3DS again if necessary

alexbishop1:
When frontend calls connector to authorise a 3D Secure result with the payment gateway, redirect the user to the 3DS Secure page again if connector’s response indicates that this is required.

Connector indicates this by sending a response body of `{"status": "AUTHORISATION 3DS REQUIRED"}` when `/v1/frontend/charges/{chargeId}/3ds` is called (this isn’t actually implemented in connector yet but will be soon).

with @SandorArpa

### DIFF
--- a/app/controllers/three_d_secure_controller.js
+++ b/app/controllers/three_d_secure_controller.js
@@ -52,7 +52,12 @@ const handleThreeDsResponse = (req, res, charge) => response => {
   switch (response.statusCode) {
     case 200:
     case 400:
-      redirect(res).toConfirm(charge.id)
+      if (_.get(response, 'body.status', '') === 'AUTHORISATION 3DS REQUIRED') {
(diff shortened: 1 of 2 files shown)


app/controllers/three_d_secure_controller.js
```diff
@@ -52,7 +52,12 @@ const handleThreeDsResponse = (req, res, charge) => response => {
   switch (response.statusCode) {
     case 200:
     case 400:
-      redirect(res).toConfirm(charge.id)
+      logger.info('3DS required again', getLoggingFields(req))
```
sfount: It looks like this log should only be called inside of the first conditional?

alexbishop1: Whoops!

PR #1443, opened 2020-03-18
PP-6230 Accessibility Issue with Email

iqbalgds:
* Add 'word-break: break-all' CSS to break long e-mails onto multiple lines

### DIFF
--- a/app/views/charge.njk
+++ b/app/views/charge.njk
@@ -385,7 +385,7 @@
         </div>
         {% endif %}
         {% if (gatewayAccount.emailCollectionMode === 'MANDATORY') or (gatewayAccount.emailCollectionMode === 'OPTIONAL') %}
-        <div class="govuk-!-width-three-quarters govuk-!-padding-top-4 govuk-!-margin-top-8 pay-!-border-top">
+        <div class="email-container govuk-!-width-three-quarters govuk-!-padding-top-4 govuk-!-margin-top-8 pay-!-border-top">
             <div class="govuk-form-group {% if highlightErrorFields.email %} govuk-form-group--error{% endif %}" data-validation="email">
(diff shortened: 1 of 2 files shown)


app/views/charge.njk
```diff
@@ -385,7 +385,7 @@
         </div>
         {% endif %}
         {% if (gatewayAccount.emailCollectionMode === 'MANDATORY') or (gatewayAccount.emailCollectionMode === 'OPTIONAL') %}
-        <div class="govuk-!-width-three-quarters govuk-!-padding-top-4 govuk-!-margin-top-8 pay-!-border-top">
+        <div class="email-container govuk-!-width-three-quarters govuk-!-padding-top-4 govuk-!-margin-top-8 pay-!-border-top">
```
alexbishop1: Could this be more narrowly applied?

I assume it’s for the “An email will be sent to:” part:

image

Could this be achieved by applying the style to the (dynamically-generated) `div` with the `id` of `email-confirmation`?

alexbishop1: Fine to apply it to a class but my point was that I think it could possibly be applied to an element that just contains the “An email will be sent to:” content, rather than an ancestor element that covers the whole email part of the form.

iqbalgds: That's what I wanted to do originally.
But that box is added with JS and it would mean updating the JS function to include the CSS class.  
As the function is a generic function, i.e. not just for the email field, I decided not to do that.
 I thought it would be safer to just add the class to the container.

alexbishop1: So would using the ID as the selector be that terrible in that case?

iqbalgds: Just really don't like using Id selectors in CSS as it can cause problems later on.

PR #1453, opened 2020-03-23
PP-6240 Log common line when rendering error page

nimalank7:
Description:
- We want to be able to search in Splunk for when an error page has rendered and the reason why it has
- Logging message and error has been moved into the error response methods

Co-authored-by: nimalank7 <nimalan.kirubakaran@gmail.com>

### DIFF
--- a/app/controllers/charge_controller.js
+++ b/app/controllers/charge_controller.js
@@ -92,9 +92,8 @@ const handleCreateResponse = (req, res, charge, response) => {
       }
       break
     case 500:
-      logging.failedChargePost(409, getLoggingFields(req))
-      logger.error('Charge create error response', getLoggingFields(req))
-      responseRouter.response(req, res, 'SYSTEM_ERROR', withAnalytics(charge, { returnUrl: routeFor('return', charge.id) }))
+      logging.failedChargePost(500, getLoggingFields(req))
+      responseRouter.systemErrorResponse(req, res, '500 response when authorising charge', withAnalytics(charge, { returnUrl: routeFor('return', charge.id) }))
       break
     default:
       redirect(res).toNew(req.chargeId)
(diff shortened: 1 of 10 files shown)


app/utils/response_router.js
```diff
@@ -234,7 +234,27 @@ const actions = {
   }
 }
 
-exports.response = (req, res, actionName, options) => {
+exports.errorResponse = function errorReponse (req, res, reason, options, error) {
+  logger.error('Rendering error response', {
+    page: 'ERROR',
+    reason: reason,
```
sfount: Very minor but these can just be
```
{
  reason,
  error
}
```

when the key/ value pairs match

app/controllers/charge_controller.js
```diff
@@ -159,7 +157,7 @@ module.exports = {
       try {
         await Charge(req.headers[CORRELATION_HEADER]).patch(req.chargeId, 'replace', 'email', userEmail, getLoggingFields(req))
       } catch (err) {
-        return responseRouter.response(req, res, 'SYSTEM_ERROR', withAnalytics(charge))
+        return responseRouter.systemErrorResponse(req, res, 'Error patching email address on Charge', withAnalytics(charge), err)
```
sfount: so much more information 👏

stephencdaly:
> This looks like a fantastic refactor, so much information is lost in the logging steps currently. Great work 👏
> 
> With the number of places we're passing `req` and `res` it might be useful to have some targeted integration tests to make sure that these error responses call all of the right methods and end up responded -- this should mostly be covered by existing tests and hopefully the whole pattern will be refactored away at some point.

In absence of being able to trust that there are existing tests for all these cases, we picked a number where there were cypress tests for the success scenarios and modified some stubs locally to ensure that the error views were being rendered correctly, and relied on visual analysis to reassure us that we hadn't broken anything.

Do you think it's worthwhile us going through and adding integration tests for all scenarios, or were you just suggesting there is at least one or two to cover some places where we can render an error response?

app/utils/response_router.js
stephencdaly: 👍 done

sfount:
No certainly not for all scenarios, just for the logging utility itself (i.e one error response, one system error response). It looks like this change has been carefully verified though, I think if there is an additional place we would like to cover it could be done in a separate PR.

PR #1462, opened 2020-03-27
PP-6240 Log all times we show error page

stephencdaly:
Log all times we show either the "An error occurred" or
"We’re experiencing technical problems" pages.

This is to drive an SLI that will cover all times we display and error
to the user that will probably mean they can't complete their payment.

### DIFF
--- a/app/utils/response_router.js
+++ b/app/utils/response_router.js
@@ -28,7 +28,7 @@ const systemError = {
   view: 'errors/system_error',
   analyticsPage: '/error',
   terminal: true,
-  logErrorPageShown: true
+  shouldLogErrorPageShown: true
 }
 
 const error = {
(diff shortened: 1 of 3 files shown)


app/utils/response_router.js
```diff
@@ -234,35 +239,47 @@ const actions = {
...
 }
 
 exports.response = function response (req, res, actionName, options) {
   options = options || {}
   options.viewName = actionName
   let action = lodash.result(actions, actionName)
+
+  if (action && action.logErrorPageShown) {
+    logErrorPageShown(action.view, `Action: ${actionName}`, getLoggingFields(req))
+  }
```
sfount: Would it make sense for all of this logic to go in the `render` method itself? That way none of the higher level functions would have to call this logging directly. 

It could be removed from `repsonse`, `systemErrorResponse` and `errorResponse` and let the logic not branch in multiple places.

PR #1491, opened 2020-04-21
PP-6328: Add logging if GOOGLE_PAY_MERCHANT_ID_2 is being used

oswaldquek:
This will help give us confidence during testing that a payment is using
GOOGLE_PAY_MERCHANT_ID_2.

### DIFF
--- a/app/controllers/charge_controller.js
+++ b/app/controllers/charge_controller.js
@@ -8,7 +8,8 @@ const i18n = require('i18n')
 const {
   WORLDPAY_3DS_FLEX_DDC_TEST_URL,
   WORLDPAY_3DS_FLEX_DDC_LIVE_URL,
-  DECRYPT_AND_OMIT_CARD_DATA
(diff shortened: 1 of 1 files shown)


app/controllers/charge_controller.js
```diff
@@ -50,9 +51,15 @@ const appendChargeForNewView = async function appendChargeForNewView(charge, req
   charge.allowApplePay = charge.gatewayAccount.allowApplePay
   charge.allowGooglePay = charge.gatewayAccount.allowGooglePay
   charge.googlePayGatewayMerchantID = charge.gatewayAccount.gatewayMerchantId
+  
+  const merchantId = getMerchantId(charge.gatewayAccount.gatewayAccountId)
```
alexbishop1, line 55: Can this maybe be `googlePayMerchantId`? On its own, `merchantId` is a bit non-specific.

PR #1503, opened 2020-04-23
PP-6361 Send ePDQ 3DS2 data to connector

alexbishop1:
Description:
- This PR sends the extra browser details required for ePDQ 3DS2 to Connector

with @nimalank7

### DIFF
--- a/test/integration/card_details_send_epdq_3ds2_data_ft_tests.js
+++ b/test/integration/card_details_send_epdq_3ds2_data_ft_tests.js
@@ -1,0 +1,181 @@
+'use strict'
+
+// NPM dependencies
+const _ = require('lodash')
+const nock = require('nock')
+const logger = require('../../app/utils/logger')(__filename)
+const proxyquire = require('proxyquire')
+
+// Local dependencies
+const cookie = require('../test_helpers/session')
+const helper = require('../test_helpers/test_helpers')
+const {
+  postChargeRequest,
+  defaultConnectorResponseForGetCharge,
+  defaultAdminusersResponseForGetService
+} = helper
+const State = require('../../config/state')
+
+// Constants
+const app = proxyquire('../../server.js',
+  {
+    'memory-cache': {
+      get: function () {
+        return false
+      },
+      '@global': true
+    }
+  }).getApp()
+
+let mockServer
+
+const gatewayAccount = {
+  gatewayAccountId: '12345',
+  paymentProvider: 'epdq',
+  analyticsId: 'test-1234',
+  type: 'test'
+}
+
+const chargeId = '23144323'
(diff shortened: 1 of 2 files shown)


test/integration/card_details_send_epdq_3ds2_data_ft_tests.js
```diff
@@ -0,0 +1,189 @@
...
+    const cookieValue = cookie.create(chargeId)
+
+    nock(process.env.CONNECTOR_HOST)
+      .patch('/v1/frontend/charges/' + chargeId)
+      .reply(200)
+    defaultConnectorResponseForGetCharge(chargeId, State.ENTERING_CARD_DETAILS, gatewayAccountId)
+    defaultAdminusersResponseForGetService(gatewayAccountId)
+    mockSuccessCardIdResponse()
+
+    const dataSentToConnector = expectedDataSentToConnector
```
oswaldquek, line 154: does this reassignment need to happen?

alexbishop1: Not really — will change

test/integration/card_details_send_epdq_3ds2_data_ft_tests.js
```diff
@@ -0,0 +1,189 @@
...
```
oswaldquek, line 176: I think this assignment doesn't need to happen either. I did a simple test:
```
> var list1 = [1,2,3]
undefined
> var list2 = list1
undefined
> list2[1] = 5
5
> list1
[ 1, 5, 3 ]
> list2
[ 1, 5, 3 ]
```
Reassigning it doesn't create a new object assigned to a new variable. The original object here `expectedDataSentToConnector` will change when `dataSentToConnector` is changed.

test/integration/card_details_send_epdq_3ds2_data_ft_tests.js
```diff
@@ -0,0 +1,189 @@
...
+    delete dataSentToConnector.js_screen_color_depth
+
+    connectorExpects(dataSentToConnector).reply(200)
+
+    const formData = formSubmissionData
```
oswaldquek, line 181: same comment as above

test/integration/card_details_send_epdq_3ds2_data_ft_tests.js
alexbishop1: Yeah, the intention wasn’t to clone the object.

Honestly, some of the code in this test was ~copied~ inspired by other tests and in the other tests the equivalent of `expectedDataSentToConnector` was a function rather than a variable (it took some arguments and returned an appropriate object), so the whole assignment thing is just an artefact of that. Will remove.

PR #1510, opened 2020-04-28
PP-6380 - Log more information for Worldpay 3DS

nimalank7:
Description:
- We want to log the information we send to the card issuer as well as Worldpay so that we can diagnose 3DS retries.

### DIFF
--- a/app/controllers/three_d_secure_controller.js
+++ b/app/controllers/three_d_secure_controller.js
@@ -31,7 +31,7 @@ const build3dsPayload = (chargeId, req) => {
   const paRes = _.get(req, 'body.PaRes')
   if (!_.isUndefined(paRes)) {
     auth3dsPayload.pa_response = paRes
-    logger.info(`paRes received for charge [${chargeId}] 3DS authorisation [starts with [${paRes.substring(0, 50)}] and ending with [${paRes.substring(paRes.length - 50)}] ]`,
+    logger.info(`paRes received for charge [${chargeId}] 3DS authorisation [${paRes}]`,
       getLoggingFields(req))
   }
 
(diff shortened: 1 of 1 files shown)


app/controllers/three_d_secure_controller.js
```diff
@@ -98,16 +98,26 @@ module.exports = {
...
         postUrl: issuerUrl,
         paRequest: paRequest,
-        threeDSReturnUrl: `${req.protocol}://${req.hostname}${paths.generateRoute('external.card.auth3dsRequiredIn', { chargeId: charge.id })}`
+        threeDSReturnUrl: threeDSReturnUrl
       }
       if (md) {
         data.md = md
       }
+
+      logger.info('Form details sent to card issuer', {
```
alexbishop1, line 113: The form hasn’t actually been submitted yet (it probably will be but something could go wrong) and it would probably be useful to specifically name the form we’re loading, so maybe this could be something like: “Rendering form to posts to card issuer or gateway to initiate 3DS authentication”.

app/controllers/three_d_secure_controller.js
```diff
@@ -98,16 +98,26 @@ module.exports = {
...
+        ...getLoggingFields(req),
+        post_url: issuerUrl,
+        pa_request: paRequest,
```
alexbishop1, line 116: I think we probably want to use the actual names of the form fields rather than the names of the variables we put in them (so `PaReq` rather than `paRequest`) in order to make it easier to reconstruct what will actually be posted.

There’s also the issue of that the form fields will get jumbled up with all of the other fields we log. I haven’t actually tried this but I assume we could maybe use nested objects to make this clear. Something like this:

```js
logger.info('Rendering form to posts to card issuer or gateway to initiate 3DS authentication',
  ...getLoggingFields(req),
  form: {
    action: issuerUrl,
    fields: {
      PaReq: paRequest,
      MD: md,
      TermUrl: threeDSReturnUrl
    }
  }
```

Might be worth trying that and seeing what happens.

app/controllers/three_d_secure_controller.js
```diff
@@ -98,16 +98,30 @@ module.exports = {
...
         postUrl: issuerUrl,
         paRequest: paRequest,
-        threeDSReturnUrl: `${req.protocol}://${req.hostname}${paths.generateRoute('external.card.auth3dsRequiredIn', { chargeId: charge.id })}`
+        threeDSReturnUrl: threeDSReturnUrl
       }
       if (md) {
         data.md = md
       }
+
+      logger.info('Rendering form to post to card issuer', {
```
alexbishop1, line 113: Something like “Rendering form to post to card issuer to initiate 3DS authentication” would be a bit more precise, just to add some more context.

app/controllers/three_d_secure_controller.js
```diff
@@ -117,6 +131,19 @@ module.exports = {
         postUrl: challengeUrl,
         worldpayChallengeJwt: worldpayChallengeJwt
       }
+
+      logger.info('Rendering form to post to card issuer', {
```
alexbishop1, line 135: In this case we know specifically that it’s a Worldpay 3DS Flex payment and therefore posting to Worldpay, so something like “Rendering form to post to Worldpay to initiate 3DS authentication” would be good.

app/controllers/three_d_secure_controller.js
nimalank7: Done

app/controllers/three_d_secure_controller.js
nimalank7: Done

app/controllers/three_d_secure_controller.js
nimalank7: Done

app/controllers/three_d_secure_controller.js
nimalank7: Done

PR #1610, opened 2020-06-16
ZD#4115502 Fix total amount calculation

kbottla:
## WHAT
- Consider pence for total amount calculation when applying corporate surchage on `entering card details page`
- Currently amount (from enter card details page) is treated as integer which is resulting in incorrect total amount on the page. There is no impact on the backend


kbottla:
Thanks @alexbishop1. Can't see tests for form validation scripts, but there may be cypress tests. Will have a look and update if there are any or will add a story to address it.

PR #1635, opened 2020-06-26
PP-6611 Fix Apple Pay when email collection is off

alexbishop1:
Fix Apple Pay when email collection is off by making it not try to validate an email address that isn’t there, which causes an error, which causes the Apple Pay authorisation process to hang for an uncomfortably long time before failing with an error.

### DIFF
--- a/test/fixtures/wallet_payment_fixtures.js
+++ b/test/fixtures/wallet_payment_fixtures.js
@@ -58,6 +58,9 @@ const fixtures = {
         }
       }
     }
+    if (ops.email) {
+      data.payment_info.email = ops.email
+    }
     return {
       getPactified: () => {
         return pactBase.pactify(data)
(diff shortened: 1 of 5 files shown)


test/fixtures/wallet_payment_fixtures.js
```diff
@@ -58,6 +58,9 @@ const fixtures = {
         }
       }
     }
+    if (ops.email) {
```
danworth, line 61: Is it still intended that the email defaults to `name@email.fyi` as per line 48 (I think the intention is that no email is sent if not sent in the opts)?
```
payment_info: {
        last_digits_card_number: ops.lastDigitsCardNumber || successfulLastDigitsCardNumber,
        brand: 'master-card',
        cardholder_name: 'Some Name',
        email: 'name@email.fyi'
      },
```

alexbishop1: No, you’re right, I missed that!

alexbishop1: PR to fix that: https://github.com/alphagov/pay-frontend/pull/1655

PR #1640, opened 2020-06-29
PP-6706 Make example card expiry date always be in the future

alexbishop1:
Make the example card expiry date on the enter card details page always use a year two years ahead of the current year in the UK.

When it’s 2020 in the UK, the example card expiry date will be 10/22. When it’s 2021 in the UK, the example card expiry date will be 10/23. When it’s 2022 in the UK, the example card expiry date will be 10/24. And so on. (The month always remains 10.)

Expiry date

This means that the example card expiry date will be between 1 year 10 months and 2 years 10 months in the future, which is appropriate given that most cards expire between 3 and 5 years after they are issued.

The example year will change just once each year at the midnight between New Year’s Eve and New Year’s Day in the UK.

### DIFF
--- a/app/controllers/charge_controller.js
+++ b/app/controllers/charge_controller.js
@@ -20,6 +20,7 @@ const Charge = require('../models/charge')
 const Card = require('../models/card')
 const State = require('../../config/state')
 const paths = require('../paths')
+const { getFutureYearAs2Digits } = require('../services/example_card_expiry_date')
 const { countries } = require('../services/countries')
 const { commonTypos } = require('../utils/email_tools')
 const { withAnalyticsError, withAnalytics } = require('../utils/analytics')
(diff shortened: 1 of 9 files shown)


app/controllers/charge_controller.js
```diff
@@ -45,16 +46,17 @@ const appendChargeForNewView = async function appendChargeForNewView (charge, re
     prepaidCredit: charge.gatewayAccount.corporatePrepaidCreditCardSurchargeAmount,
     prepaidDebit: charge.gatewayAccount.corporatePrepaidDebitCardSurchargeAmount
   })
-  charge.post_card_action = routeFor('create', chargeId)
   charge.id = chargeId
+  charge.exampleCardExpiryDateYear = getTwoDigitYearInFuture()
```
sfount, line 50: GitHub discarded my previous comment and I've lost the will so less details now 🤦 .

We're conflating the view context and the charge (already done before this PR, this file is some of the most confusing code I've ever read!), assigning this payment page specific value to something called `charge` is confusing. 

This information could be passed to the view context at the point of rendering, unfortunately that happens 3 times in this controller because of how it's organised. 

It would look something like:
```js
responseRouter.response(req, res, views.CHARGE_VIEW, withAnalytics(charge, { charge, exampleCardExpiryDateYear })
```

Also `withAnalytics(charge, charge)` 🤨 

As I mentioned this PR doesn't cause this behaviour so it shouldn't be required to address this but the data definitely doesn't belong to the charge.

app/services/example_card_expiry_date.js
```diff
@@ -0,0 +1,11 @@
+'use strict'
+
+const moment = require('moment-timezone')
+
+const YEARS_TO_ADD = 2
+
+const getTwoDigitYearInFuture = () => moment(new Date()).tz('Europe/London').add(YEARS_TO_ADD, 'years').format('YY')
```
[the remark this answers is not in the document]
alexbishop1: I can see how `getTwoDigitYearInFuture` could be misinterpreted: I’ll rename it to `getFutureYearAs2Digits`, which will hopefully be clearer.

I almost didn’t have the `YEARS_TO_ADD` constant: I only added it due to a fear of magic numbers. I think I’ll just remove it.

Not too persuaded by having a `yearsToAdd` parameter. The Cypress test will break if it’s anything other than 2 and I don’t think this is something we‘re going to be changing very often.

app/controllers/charge_controller.js
alexbishop1: Oh yes, I felt terrible adding `exampleCardExpiryDateYear` to the `charge` but did it because that was the existing pattern. However, while most of the properties `charge` is burdened with can vary between payments (and therefore be considered to belong to the `charge` if you squint a bit), there’s nothing at all payment-specific about the example expiry date. I’ll have a look at doing what you suggest.

alexbishop1: I did start doing this but it seemed to have way more ripple effects than I anticipated, so I think we’ll have to leave it for now.

PR #1644, opened 2020-07-01
BAU add default label for dependabot updates

SandorArpa:
## WHAT
- add govuk-pay default label to repo so it shows up in the unified dependabot updates view


SandorArpa:
Failed to open ci build

PR #1665, opened 2020-07-14
PP-6613 - Allow all supported cards as normalised for Apple Pay

nimalank7:
Description:
- Apple Pay code which filters the supported card brands allows Discover, JCB, Maestro and Visa Electron for Apple Pay. However
when we normalise the card brand we only allow Visa, Mastercard or American Express which causes an error for users who try to pay with
Discover, JCB, Maestro or Visa Electron.
- This PR modifies the normalisation code to allow for these.
- A test has been added to check to see whether an exception is thrown when an unrecognizable card brand has been inputted. We decided against testing all the cards as testing a single card would be sufficient.

Paired with @narinderchana

### DIFF
--- a/test/controllers/web-payments/apple-pay/normalise-apple-pay-payload.test.js
+++ b/test/controllers/web-payments/apple-pay/normalise-apple-pay-payload.test.js
@@ -53,6 +53,36 @@ describe('normalise apple pay payload', function () {
       }
     )
   })
+  it('should return an error when an unrecognizable card brand is passed in', function () {
+    const applePayPayload = {
+      shippingContact: {
+        emailAddress: 'jonheslop@bla.test',
+        familyName: 'payment',
+        givenName: 'mr',
+        phoneticFamilyName: '',
+        phoneticGivenName: ''
+      },
+      token: {
+        paymentData: {
+          version: 'EC_v1',
+          data: 'MLHhOn2BXhNw9wLLDR48DyeUcuSmRJ6KnAIGTMGqsgiMpc+AoJ…LUQ6UovkfSnW0sFH6NGZ0jhoap6LYnThYb9WT6yKfEm/rDhM=',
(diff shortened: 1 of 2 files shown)


test/controllers/web-payments/apple-pay/normalise-apple-pay-payload.test.js
```diff
@@ -53,6 +53,36 @@ describe('normalise apple pay payload', function () {
...
+          signature: 'MIAGCSqGSIb3DQEHAqCAMIACAQExDzANBglghkgBZQMEAgEFAD…ZuQFfsLJ+Nb3+7bpjfBsZAhA1sIT1XmHoGFdoCUT3AAAAAAAA',
+          header: {
+            ephemeralPublicKey: 'MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE5/Qc6z4TY5HQ5n…KC3kJ4DtIWedPQ70N35PBZzJUrFjtvDZFUvs80uo2ynu+lw==',
+            publicKeyHash: 'Xzn7W3vsrlKlb0QvUAviASubdtW4BotWrDo5mGG+UWY=',
+            transactionId: '372c3858122b6bc39c6095eca2f994a8aa012f3b025d0d72ecfd449c2a5877f9'
+          }
+        },
+        paymentMethod: {
+          displayName: 'Undefined',
+          network: 'undefined',
```
sfount, line 78: Probably very minor! If possible in this test it might be a good idea to use anything other than `null` or `undefined` for the input. 

JavaScript does some funky things with type coercion (ref: `==` `===`) and might happily take a scenario where the thing is actually not defined and throw `undefined` in a string which would match the assertion. 

I may not have understood this properly but any other string should work well here and give a definite assertion.

nimalank7: Yeah good shout!

PR #1672, opened 2020-07-16
PP-6614 Normalise the last four digits for Apple Pay

nimalank7:
- If the displayName is less than four characters then map the last four digits to an empty string.
- If the last four characters of the displayName are not numeric then map the last four digits to an empty string.

Paired with @narinderchana

### DIFF
--- a/app/controllers/web-payments/apple-pay/normalise-apple-pay-payload.js
+++ b/app/controllers/web-payments/apple-pay/normalise-apple-pay-payload.js
@@ -3,6 +3,7 @@
 const logger = require('../../../utils/logger')(__filename)
 const { getLoggingFields } = require('../../../utils/logging_fields_helper')
 const { output, redact } = require('../../../utils/structured_logging_value_helper')
+const isNumeric = require('../../../utils/number')
 const humps = require('humps')
 
 const logselectedPayloadProperties = req => {
@@ -80,7 +81,14 @@ const nullable = word => {
   return word
 }
 
-const normaliseLastDigitsCardNumber = displayName => displayName.substr(displayName.length - 4)
+const normaliseLastDigitsCardNumber = displayName => {
+  let lastDigitsCardNumber = ''
+  const lastFourChars = displayName.substr(displayName.length - 4)
+  if (displayName.length >= 4 && isNumeric(lastFourChars)) {
+    lastDigitsCardNumber = lastFourChars
+  }
+  return lastDigitsCardNumber
+}
 
 module.exports = req => {
   logselectedPayloadProperties(req)
--- a/app/utils/number.js
+++ b/app/utils/number.js
@@ -1,0 +1,5 @@
+'use strict'
+
+module.exports = str => {
+  return !isNaN(str) && !isNaN(parseInt(str))
+}
--- a/test/controllers/web-payments/apple-pay/normalise-apple-pay-payload.test.js
+++ b/test/controllers/web-payments/apple-pay/normalise-apple-pay-payload.test.js
@@ -53,6 +53,124 @@ describe('normalise apple pay payload', function () {
       }
     )
   })
+
+  it('should return an empty string for last_digits_card_number when displayName does not have numeric values for last 4 characters', function () {
+    const applePayPayload = {
+      shippingContact: {
+        emailAddress: 'jonheslop@bla.test',
+        familyName: 'payment',
+        givenName: 'mr',
+        phoneticFamilyName: '',
+        phoneticGivenName: ''
+      },
+      token: {
+        paymentData: {
+          version: 'EC_v1',
+          data: 'MLHhOn2BXhNw9wLLDR48DyeUcuSmRJ6KnAIGTMGqsgiMpc+AoJ…LUQ6UovkfSnW0sFH6NGZ0jhoap6LYnThYb9WT6yKfEm/rDhM=',
+          signature: 'MIAGCSqGSIb3DQEHAqCAMIACAQExDzANBglghkgBZQMEAgEFAD…ZuQFfsLJ+Nb3+7bpjfBsZAhA1sIT1XmHoGFdoCUT3AAAAAAAA',
+          header: {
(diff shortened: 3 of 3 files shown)


app/controllers/web-payments/apple-pay/normalise-apple-pay-payload.js
```diff
@@ -80,7 +81,14 @@ const nullable = word => {
   return word
 }
 
-const normaliseLastDigitsCardNumber = displayName => displayName.substr(displayName.length - 4)
+const normaliseLastDigitsCardNumber = displayName => {
+  let lastDigitsCardNumber = ''
+  const lastFourChars = displayName.substr(displayName.length - 4)
```
alexbishop1, line 86: You can actually pass a negative number to `substr(…)` to have it start counting from the end of the string rather than the beginning, so `displayName.substr(-4)` would be equivalent if `displayName` is less than 4 characters long.

I think this be easier to understand if the substring was taken after the `displayName.length >= 4` check, so the reader doesn’t have to try to figure out what  `displayName.substr(-4)` will do if `displayName` is under four characters long.

app/utils/number.js
```diff
@@ -0,0 +1,5 @@
+'use strict'
+
+module.exports = str => {
+  return !isNaN(str) && !isNaN(parseInt(str))
```
alexbishop1, line 4: The [`parseInt(…)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt) function will parse things that aren’t Arabic numerals (for example, it will allow hexadecimal).

Rather than trying to parse and integer, it’s probably easier to just use a regex:

```javascript
if (displayName.match(/.*[0-9]{4}$/)) {
  // …
}
```

test/controllers/web-payments/apple-pay/normalise-apple-pay-payload.test.js
```diff
@@ -53,6 +53,124 @@ describe('normalise apple pay payload', function () {
...
+          data: 'MLHhOn2BXhNw9wLLDR48DyeUcuSmRJ6KnAIGTMGqsgiMpc+AoJ…LUQ6UovkfSnW0sFH6NGZ0jhoap6LYnThYb9WT6yKfEm/rDhM=',
+          signature: 'MIAGCSqGSIb3DQEHAqCAMIACAQExDzANBglghkgBZQMEAgEFAD…ZuQFfsLJ+Nb3+7bpjfBsZAhA1sIT1XmHoGFdoCUT3AAAAAAAA',
+          header: {
+            ephemeralPublicKey: 'MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE5/Qc6z4TY5HQ5n…KC3kJ4DtIWedPQ70N35PBZzJUrFjtvDZFUvs80uo2ynu+lw==',
+            publicKeyHash: 'Xzn7W3vsrlKlb0QvUAviASubdtW4BotWrDo5mGG+UWY=',
+            transactionId: '372c3858122b6bc39c6095eca2f994a8aa012f3b025d0d72ecfd449c2a5877f9'
+          }
+        },
+        paymentMethod: {
+          displayName: 'JBC',
```
alexbishop1, line 117: Making this be a string that’s longer than 4 characters but only ends in 3 digits (e.g. `Visa 123`) would be a more rigorous test.

test/controllers/web-payments/apple-pay/normalise-apple-pay-payload.test.js
```diff
@@ -53,6 +53,124 @@ describe('normalise apple pay payload', function () {
...
+          data: 'MLHhOn2BXhNw9wLLDR48DyeUcuSmRJ6KnAIGTMGqsgiMpc+AoJ…LUQ6UovkfSnW0sFH6NGZ0jhoap6LYnThYb9WT6yKfEm/rDhM=',
+          signature: 'MIAGCSqGSIb3DQEHAqCAMIACAQExDzANBglghkgBZQMEAgEFAD…ZuQFfsLJ+Nb3+7bpjfBsZAhA1sIT1XmHoGFdoCUT3AAAAAAAA',
+          header: {
+            ephemeralPublicKey: 'MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE5/Qc6z4TY5HQ5n…KC3kJ4DtIWedPQ70N35PBZzJUrFjtvDZFUvs80uo2ynu+lw==',
+            publicKeyHash: 'Xzn7W3vsrlKlb0QvUAviASubdtW4BotWrDo5mGG+UWY=',
+            transactionId: '372c3858122b6bc39c6095eca2f994a8aa012f3b025d0d72ecfd449c2a5877f9'
+          }
+        },
+        paymentMethod: {
+          displayName: '       ',
```
alexbishop1, line 156: This is a blank string, not an empty string! 😀

narinderchana: Point taken about the blank string, we will rename the test it's blank string.

We will incorporate the substr, condition handling and regex in the next branch commit

Thank you

test/controllers/web-payments/apple-pay/normalise-apple-pay-payload.test.js
narinderchana: We have to keep it as JBC because this test is for less than four chars, other we won't be test that condition.

such as displayName.length < 4

app/controllers/web-payments/apple-pay/normalise-apple-pay-payload.js
narinderchana: We are going do the Pact tests in another PR.

test/controllers/web-payments/apple-pay/normalise-apple-pay-payload.test.js
alexbishop1: Oh, right. In that case, maybe make it `123` so we know we’re definitely hitting the length check and not the is-it-digits check.

PR #1673, opened 2020-07-16
BAU Extend 3DS Flex Timeout

rorymalcolm:
- Extend the timeout of 3DS Flex queries to CC to increase potential
resolution and instances of the 3DS Flex check as opposed to a default
behaviour fallback.

### DIFF
--- a/app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
+++ b/app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
@@ -43,9 +43,11 @@ const checkOriginHostName = origin => {
 }
 
 const submitWithWorldpay3dsFlexDdcResult = form => {
+  const DDC_TIMEOUT_IN_MILLISECONDS = 30000
+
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
```diff
@@ -43,9 +43,11 @@ const checkOriginHostName = origin => {
 }
 
 const submitWithWorldpay3dsFlexDdcResult = form => {
+  const WORLDPAY_RESPONSE_TIMEOUT = 30000
```
alexbishop1, line 46: `DDC_TIMEOUT_IN_MILLISECONDS`? (DDC is more specific than Worldpay — most of the time we’ll be waiting for CardinalCommerce rather than Worldpay — and including the units is always good.)

rorymalcolm: Agreed

PR #1692, opened 2020-07-23
BAU bump product pages from 4.4.2 to 4.4.3

kbottla:
## WHAT
 - updates product pages to latest release
 - uses https urls for npm repos in package-lock.json


sfount:
The product page release `4.4.3` seems to be larger than previous releases, this could just be because some additional files got included by accident but it would be good to double check the changes behaviour in test before releasing.

PR #1737, opened 2020-08-24
Revert "Revert "PP-4733: Spike sending metric to HG""

oswaldquek:
Reverts alphagov/pay-frontend#1736.


narinderchana (approved):
Looks good rolling back in the HG metircs

PR #1738, opened 2020-08-24
Revert "Revert "Revert "PP-4733: Spike sending metric to HG"""

oswaldquek:
Reverts alphagov/pay-frontend#1737


narinderchana (approved):
Rolling back the HG, spike

PR #1759, opened 2020-09-07
PP-6995: Add graphite reporter

oswaldquek:
Add graphite reporter. Only counters are envisaged for now.


app/services/clients/connector_client.js
```diff
@@ -187,11 +191,16 @@ const _getConnector = (url, description, loggingFields = {}) => {
         url: url,
         error: err
       })
+      getCounter(`${METRICS_PREFIX}.${callingFunction}.error`).inc()
```
oswaldquek, line 194: As per the story the idea here was to log timeouts but I can't see how we can know when an `err` is a timeout

narinderchana (approved):
Looks good

PR #1760, opened 2020-09-07
PP-7075 Add logging for DDC request for Worldpay 3DS flex

stephencdaly:
Add a hidden form field to the payment details form containing a status for the Device Data Collection (DDC) for Worldpay 3DS flex.

Update this status at various points in performing the DDC request. On the server side, log the value of this hidden status field so we can use this to determine at what point the DDC went wrong if we do not get a result to use.

Regenerate .secrets.baseline to ignore package.lock.json

### DIFF
--- a/app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
+++ b/app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
@@ -42,10 +42,15 @@ const checkOriginHostName = origin => {
   return Charge.worldpay_3ds_flex_ddc_url.lastIndexOf(origin, 0) === 0
 }
 
+const updateDdcStatusField = (status) => {
+  document.getElementById('worldpay3dsFlexDdcStatus').value = status
+}
+
 const submitWithWorldpay3dsFlexDdcResult = form => {
   const DDC_TIMEOUT_IN_MILLISECONDS = 30000
 
   const worldpayNoResponseTimeout = window.setTimeout(() => {
+    updateDdcStatusField(`DDC timeout after ${DDC_TIMEOUT_IN_MILLISECONDS} milliseconds`)
     form.submit()
   }, DDC_TIMEOUT_IN_MILLISECONDS)
 
@@ -54,7 +59,12 @@ const submitWithWorldpay3dsFlexDdcResult = form => {
       const { MessageType, SessionId, Status } = JSON.parse(event.data)
       if (MessageType === 'profile.completed') {
         if (Status === true && typeof SessionId === 'string') {
+          updateDdcStatusField('valid DDC result')
           addWorldpaySessionIdToForm(form, SessionId)
+        } else if (!Status) {
+          updateDdcStatusField('DDC result did not have Status of true')
+        } else {
+          updateDdcStatusField('no SessionID string in DDC result')
         }
         window.clearTimeout(worldpayNoResponseTimeout)
         form.submit()
@@ -72,6 +82,7 @@ const submitWithWorldpay3dsFlexDdcResult = form => {
     iFrameContent.getElementById('input-jwt').value = Charge.worldpay_3ds_flex_ddc_jwt
     iFrameContent.getElementById('input-bin').value = form.elements.cardNo.value
     innerForm.submit()
+    updateDdcStatusField('DDC initiated')
   }
 
   if (innerDoc.readyState === 'complete') {
--- a/app/controllers/charge_controller.js
+++ b/app/controllers/charge_controller.js
@@ -72,7 +72,7 @@ const appendChargeForNewView = async function appendChargeForNewView (charge, re
   charge.worldpay3dsFlexDdcUrl = charge.gatewayAccount.type !== 'live' ? WORLDPAY_3DS_FLEX_DDC_TEST_URL : WORLDPAY_3DS_FLEX_DDC_LIVE_URL
 
   charge.collectAdditionalBrowserDataForEpdq3ds = charge.gatewayAccount.paymentProvider === 'epdq' &&
-      charge.gatewayAccount.requires3ds && charge.gatewayAccount.integrationVersion3ds === 2
+    charge.gatewayAccount.requires3ds && charge.gatewayAccount.integrationVersion3ds === 2
 }
 
 const routeFor = (resource, chargeId) => paths.generateRoute(`card.${resource}`, { chargeId: chargeId })
@@ -137,6 +137,16 @@ module.exports = {
 
     normalise.addressLines(req.body)
     normalise.whitespace(req.body)
+
+
+    const { worldpay3dsFlexDdcStatus } = req.body
+    if (worldpay3dsFlexDdcStatus) {
+      logger.info(`Payment details submitted for a Worldpay 3DS Flex charge. DDC status is: ${worldpay3dsFlexDdcStatus}`,
+        {
+          ...getLoggingFields(req),
+          worldpay_3ds_flex_ddc_status: worldpay3dsFlexDdcStatus
+        })
+    }
 
     if (charge.status === State.AUTH_READY) return redirect(res).toAuthWaiting(req.chargeId)
     // else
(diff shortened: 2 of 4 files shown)


app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
```diff
@@ -54,7 +59,12 @@ const submitWithWorldpay3dsFlexDdcResult = form => {
       const { MessageType, SessionId, Status } = JSON.parse(event.data)
       if (MessageType === 'profile.completed') {
         if (Status === true && typeof SessionId === 'string') {
+          updateDdcStatusField('valid DDC result')
           addWorldpaySessionIdToForm(form, SessionId)
+        } else if (!Status) {
+          updateDdcStatusField('DDC response had Status of false')
```
alexbishop1, line 65: I suppose technically it could be that it’s missing or an empty string or something.

app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
```diff
@@ -54,7 +59,12 @@ const submitWithWorldpay3dsFlexDdcResult = form => {
...
+        } else {
+          updateDdcStatusField('SessionID missing in DDC response')
```
alexbishop1, line 67: Technically, this could be it exists but isn’t a string.

test/cypress/integration/worldpay-3ds-flex-ddc/payment-with-session-id.spec.js
```diff
@@ -95,32 +95,36 @@ describe('Worldpay 3ds flex card payment flow', () => {
...
-    it('If Worldpay were to respond with status=false, submitting confirmation should not include the hidden input containing the session ID', () => {
+  describe('Worldpay responds with status=false', () => {
+    it('submitting confirmation should not include the hidden input containing the session ID', () => {
       cy.task('setupStubs', [...confirmPaymentDetailsStubs, worldpay3dsFlexDdcStubFailure])
 
       cy.get('#card-details').submit().should($form => {
         const formVal = $form.first()[0].elements.worldpay3dsFlexDdcResult
         expect(formVal).to.eq(undefined)
+        const ddcStatusVal = $form.first()[0].elements.worldpay3dsFlexDdcStatus.value
+        expect(ddcStatusVal).to.eq('DDC response had Status of false')
```
alexbishop1, line 120: Nice that you added these assertions but I’m quite tempted to ruin your day by asking you to add in scenarios for the other statuses we might report.

app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
```diff
@@ -42,10 +42,15 @@ const checkOriginHostName = origin => {
   return Charge.worldpay_3ds_flex_ddc_url.lastIndexOf(origin, 0) === 0
 }
 
+const updateDdcStatusField = (status) => {
+  document.getElementById('worldpay3dsFlexDdcStatus').value = status
```
alexbishop1, line 46: Wondering if we should call this “status” given DDC thing called “status”. We’ve already used “result” too. Maybe `worldpay3dsFlexDdcReport`? Or maybe it’s fine to leave it as “status”.

alexbishop1 (approved):
This looks good. Made some suggestions but none of them are show-stoppers.

test/cypress/integration/worldpay-3ds-flex-ddc/payment-with-session-id.spec.js
stephencdaly: My opinion is that would be overkill, we don't generally test logging but as these are set by client-side javascript I wanted to ensure that it didn't break in some way we didn't notice by someone changing something that links the components together.

Also can't do an assertion for the timeout in cypress tests, as it will just look at the value at the time the form is submitted.

app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
stephencdaly: i reckon it's fine, it won't take people long to figure out what it means based on how we set it

app/controllers/charge_controller.js
```diff
@@ -138,6 +138,11 @@ module.exports = {
     normalise.addressLines(req.body)
     normalise.whitespace(req.body)
 
+    if (req.body.worldpay3dsFlexDdcStatus) {
+      logger.info(`Payment details submitted for a Worldpay 3DS Flex charge. DDC status is: ${req.body.worldpay3dsFlexDdcStatus}`,
```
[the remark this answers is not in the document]
stephencdaly: yeh sure, will do

app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
stephencdaly: true, but I reckon logging something that is almost right is fine, I imagine people will be digging into code to determine what this means when its seen anyway

alexbishop1: I suppose it could say something like “no SessionID string in DDC result” (which leads me on to another thing I’ve just noticed: inconsistent about whether we call it a “result” or a “response”).

stephencdaly:
I've updated the wording of some of the statuses, and logged it as a structured argument

app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
```diff
@@ -54,7 +59,12 @@ const submitWithWorldpay3dsFlexDdcResult = form => {
       const { MessageType, SessionId, Status } = JSON.parse(event.data)
       if (MessageType === 'profile.completed') {
         if (Status === true && typeof SessionId === 'string') {
+          updateDdcStatusField('valid DDC result')
           addWorldpaySessionIdToForm(form, SessionId)
+        } else if (!Status) {
+          updateDdcStatusField('DDC result had Status of false')
```
alexbishop1, line 65: “DDC result did not have Status of true”? That would cover it not being there at all or not being a boolean.

test/cypress/integration/worldpay-3ds-flex-ddc/payment-with-session-id.spec.js
```diff
@@ -95,32 +95,36 @@ describe('Worldpay 3ds flex card payment flow', () => {
...
+        const ddcStatusVal = $form.first()[0].elements.worldpay3dsFlexDdcStatus.value
+        expect(ddcStatusVal).to.eq('DDC result had Status of false')
       })
     })
   })
 
   setUpAndCheckCardPaymentPage()
-  describe('Secure confirmation page', () => {
-    it('Submitting confirmation when Worldpay times out after  iframe post should still submit the form but without the worldpaySessionId', () => {
+  describe('Worldpay times out', () => {
```
alexbishop1, line 126: “DDC times out” would be more precise because in real life it’s not Worldpay that does it.

alexbishop1 (changes requested):
I’m being nitpicky about the strings.

app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
stephencdaly: that's the equivalent of being false in javascript haha. I was gonna say 'DDC result has a Status that was not "true"' but i thought that made less sense haha

app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
```diff
@@ -54,7 +59,12 @@ const submitWithWorldpay3dsFlexDdcResult = form => {
       const { MessageType, SessionId, Status } = JSON.parse(event.data)
       if (MessageType === 'profile.completed') {
         if (Status === true && typeof SessionId === 'string') {
+          updateDdcStatusField('valid DDC result')
           addWorldpaySessionIdToForm(form, SessionId)
+        } else if (!Status) {
+          updateDdcStatusField('DDC result had Status that was not "true"')
```
alexbishop1, line 65: But it’s not a string!

stephencdaly: i reupdated

stephencdaly: didn't see your suggestion before

alexbishop1 (approved):

PR #1761, opened 2020-09-07
PP-6995: Send metrics for PUT/POST/PATCH calls to connector

oswaldquek:
As the PR wasn't reviewed quickly enough I've included the call from frontend to adminusers.


nimalank7:
Functionality looks good to me 👍 I'll leave it to @sfount to look at the adminusers call

app/services/clients/adminusers_client.js
```diff
@@ -4,53 +4,62 @@
...
-}
+const METRICS_PREFIX = 'internal-rest-call.adminusers'
+const SUCCESS_CODES = [200, 201, 202, 204, 206]
 
 let baseUrl
 let correlationId
 
-const findServiceBy = function findServiceBy (findOptions, loggingFields = {}) {
+/** @private */
+const _getAdminUsers = (url, description, findOptions, loggingFields = {}, callingFunction) => {
```
sfount, line 18: There's no need for this method to be anonymous, this will help when looking at the stacktrace etc.
```suggestion
const _getAdminUsers = function _getAdminUsers(url, description, findOptions, loggingFields = {}, callingFunction) {
```

app/services/clients/adminusers_client.js
```diff
@@ -4,53 +4,62 @@
...
-    baseClient.get(servicesResource, params, callbackToPromiseConverter)
-      .on('error', callbackToPromiseConverter)
+    const params = {
+      correlationId: correlationId,
+      qs: {
+        gatewayAccountId: findOptions.gatewayAccountId
+      }
+    }
+    baseClient.get(url, params, null)
+    .then(response => {
```
sfount, line 36: Now that this method is promise based there's no need for `_getAdminUsers` to return a new Promise, we found doing this can make the code harder to read and is more error prone. 

Returning this promise will behave in the same was as constructing a new promise here. 

Something like:
 
```javascript
function _getAdminUsers() {
  return baseClient.get()
    .then((respone) => {
      return response.body
    })
    .catch((error) => {
      
      // ...
      throw error
    })
} 
```

app/services/clients/adminusers_client.js
```diff
@@ -4,53 +4,62 @@
...
+      }
+    }).catch(err => {
+      requestLogger.logRequestError(context, err, loggingFields)
+      incrementFailureCounter(callingFunction, 'error')
+      reject(err)
+    })
   })
 }
 
+const incrementFailureCounter = (callingFunction, statusCode) => {
```
sfount, line 54: It seems like there's an established naming pattern here but I think we should change it up. The attribute `callingFunction` is ambiguous and sounds to me like it's expecting an actual callable function. 

I think something like `httpRequestActionName` or to follow the same theme `callingFunctionName`.

sfount (commented):
Changes look like they should work 👍 a couple of comments about code structure. 

It seems like these kind of inter-micro-service metrics/ monitoring is something that AWS X-Ray does particularly well, I'm not sure if that's been considered or is too complex for this work but just thought these patterns look familiar to the hooks required for that.

oswaldquek:
@sfount we're not doing any complex inter-micro-service metrics/monitoring here, we just want to know when a call to a downstream microservice fails and we can log that easily by sending a metric. With these metrics in hosted graphite we can then build an alert (in a future story).

app/services/clients/connector_client.js
```diff
@@ -55,7 +55,7 @@ const _getPatchUrlFor = chargeId => baseUrl + CARD_CHARGE_PATH.replace('{chargeI
 const _getWorldpay3dsFlexUrlFor = chargeId => baseUrl + WORLDPAY_3DS_FLEX_JWT_PATH.replace('{chargeId}', chargeId)
 
 /** @private */
-const _putConnector = (url, payload, description, loggingFields = {}) => {
+const _putConnector = (url, payload, description, loggingFields = {}, callingFunction) => {
```
sfount, line 58: There are ~6 anonymous functions here that are complex and should be named. Happy for that to be done separately.

sfount (approved):
Comments well addressed 👍 there are a few pieces left to clean up but I don't think that has to be done as part of this PR.

app/services/clients/connector_client.js
oswaldquek: There's some more clean up to do, I'll do this in the next PR.

PR #1762, opened 2020-09-07
PP-6957 MOTO services - allow masking of card numbers and security codes

iqbalgds:
- Update `charge` template and controller
  - When gateway account has moto_mask_card_number_input=true/false or moto_mask_card_security_code_input=true/false
  - Then change the card number or security code input to become password fields.
- Create new Cypress test

image


.secrets.baseline
```diff
@@ -1,9 +1,9 @@
 {
   "exclude": {
-    "files": null,
+    "files": "package-lock.json",
     "lines": null
   },
```
iqbalgds, line 5: I updated this file to ignore the package-lock.json file.
See this: https://github.com/alphagov/gds-pre-commit/blob/master/usage.md
```
Exclude Files
If you have files which you do not wish to run the scan against, such as a package-lock.json or go.sum which are automatically generated and have a number of high entropy strings, you can use --exclude-files *your files* to ignore these during the scan, this will be noted in .secrets.baseline to prevent you having to add this argument during future runs.
```

app/views/charge.njk
```diff
@@ -145,16 +145,24 @@
...
-                    autocomplete="cc-number"
-                    value="{{cardNo}}"
-                  />
+                      type="{{cardNumberInputType}}"
+                      inputmode="numeric"
+                      pattern="[0-9]*"
+                      name="cardNo"
+                      maxlength="26"
+                      class="govuk-input govuk-!-width-three-quarters"
+                      autocomplete="cc-number"
```
alexbishop1, line 162: In the words of the HTML Standard, `autocomplete="cc-number"` is “inappropriate for a control” with `type="password"`, so we should probably just not include the `autocomplete` attribute or use `autocomplete="off"`.

The same will apply to `autocomplete="cc-csc"` for the card security code field.

test/cypress/integration/card/moto-mask-payment.spec.js
```diff
@@ -0,0 +1,195 @@
...
+      cy.task('setupStubs', createPaymentChargeStubsNoMoto)
+      cy.visit(`/secure/${tokenId}`)
+
+      cy.location('pathname').should('eq', `/card_details/${chargeId}`)
+      cy.window().its('chargeId').should('eq', `${chargeId}`)
+    })
+
+    it('Card number & security code input elements should NOT mask digits', () => {
+      cy.get('#card-no').invoke('attr', 'type').should('contain', 'text')
+      cy.get('#cvc').invoke('attr', 'type').should('contain', 'text')
```
alexbishop1, line 65: I think we can assert on the whole value here: `should('eq', 'text')`

test/cypress/integration/card/moto-mask-payment.spec.js
```diff
@@ -0,0 +1,195 @@
...
+      cy.route('POST', `/check_card/${chargeId}`).as('checkCard')
+
+      cy.get('#card-no').type(validPayment.cardNumber)
+      cy.get('#card-no').blur()
+
+      cy.wait('@checkCard')
+      cy.get('#card-no').should('not.have.class', 'govuk-input--error')
+    })
+
+    it('should display an error when a invalid card number is entered', () => {
```
alexbishop1, line 96: Good that we have a test for this.

test/cypress/integration/card/moto-mask-payment.spec.js
```diff
@@ -0,0 +1,195 @@
...
+      cy.get('#cvc').type(validPayment.securityCode)
+      cy.get('#address-line-1').type(validPayment.addressLine1)
+      cy.get('#address-city').type(validPayment.city)
+      cy.get('#address-postcode').type(validPayment.postcode)
+      cy.get('#email').type(validPayment.email)
+    })
+  })
+
+  describe('Secure confirmation page', () => {
+    it('Submitting confirmation with valid details should redirect to confirmation page', () => {
```
alexbishop1, line 183: Good that we have a test for this.

alexbishop1 (changes requested):
This looks nice. Comments mostly relate to the `autocomplete` attribute. Might be worth experimenting to see if not including it at all versus using `autocomplete="off"` is better at persuading browsers not to save it (generally browsers will ignore `autocomplete="off"` for login forms but the fact that there’s no username field might make it behave differently).

test/cypress/integration/card/moto-mask-payment.spec.js
iqbalgds: Good spot - done

app/views/charge.njk
iqbalgds: I thought `autocomplete="off"` would be the best option - as it is more explicit.  Even though the browser may override this setting.

I have updated the `security code` field and the Cypress test.

iqbalgds:
I thought autocomplete="off" would be the best option - as it is more explicit. Even though the browser password manager may override this setting for `<input type="password">`

test/cypress/integration/card/moto-mask-payment.spec.js
```diff
@@ -0,0 +1,207 @@
...
+      allowMoto: true,
+      motoMaskCardSecurityCodeInput: true
+    })
+
+  const createPaymentChargeStubsMotoMaskCardNumberAndSecurityCode = cardPaymentStubs.buildCreatePaymentChargeStubs(
+    tokenId, chargeId, 'en', gatewayAccountId, {}, {}, {
+      allowMoto: true,
+      motoMaskCardNumberInput: true,
+      motoMaskCardSecurityCodeInput: true
+    })
```
alexbishop1, line 42: Just thought — can we have a test `allowMoto: false`, `motoMaskCardNumberInput: true` and `motoMaskCardSecurityCodeInput: true` to check it doesn’t do any masking?

alexbishop1 (changes requested):
Looks good — one suggestion for an extra test.

test/cypress/integration/card/moto-mask-payment.spec.js
iqbalgds: Good idea - just added it

test/cypress/integration/card/moto-mask-payment.spec.js
```diff
@@ -0,0 +1,232 @@
...
+
+    it('Card number & security code input elements should NOT mask digits as it is not a MOTO service', () => {
+      cy.get('#card-no').invoke('attr', 'type').should('eq', 'text')
+      cy.get('#card-no').invoke('attr', 'autocomplete').should('eq', 'cc-number')
+
+      cy.get('#cvc').invoke('attr', 'type').should('eq', 'text')
+      cy.get('#cvc').invoke('attr', 'autocomplete').should('eq', 'cc-csc')
+    })
+  })
```
iqbalgds, line 95: New test - when `MOTO=false` but `maskCardNumber=true` and `maskSecurityCode=true`

alexbishop1 (approved):
Awesomesauce!

alexbishop1 (approved):

PR #1763, opened 2020-09-08
PP-6995: Simplify connector client

oswaldquek:
* no need to return an implicit Promise
* get rid of anonymous functions


sfount (approved):
Really useful refactor 👍 if tests are happy this looks all good to go

PR #1764, opened 2020-09-08
PP-6995: Log downstream errors

oswaldquek:
The acceptance criteria on the story says 
```
Log a structured log line at error level with the message Error communicating with https://the-url-goes-here and a url key containing the URL and, if appropriate, a remote_http_status key containing the HTTP status code.
```
However some errors have already been logged with other messages. The important thing is that errors are being logged so I'm not too concerned about logging `Error communicating with https://the-url-goes-here`. 
`status_code` is already used in places so using this instead of `remote_http_status`.


nimalank7 (approved):
LGTM!

PR #1770, opened 2020-09-10
PP-6995: Log all status codes

oswaldquek:
Emit metrics for all status codes to make it easier to have confidence in what is going on.


nimalank7 (approved):
LGTM!

PR #1778, opened 2020-09-14
PP-7118 Ignore cookies for 3ds urls

narinderchana:
Description:
- We are making this change due to changes to SameSite cookie behaviour changing in some browsers (chrome) which causes the session cookie to be lost inside 3DS iframe if too long a time is spent on on the 3D secure page.
- Use the chargeId from /card_details/:chargeId/3ds_required_in and /card_details/:chargeId/3ds_required_in/epdq post urls.

### DIFF
--- a/app/services/charge_param_retriever.js
+++ b/app/services/charge_param_retriever.js
@@ -7,20 +7,25 @@ const cookie = require('../utils/cookies')
 
 exports.retrieve = req => {
   const chargeId = req.params.chargeId ? req.params.chargeId : req.body.chargeId
-  setLoggingField(req, PAYMENT_EXTERNAL_ID, chargeId)
 
+  setLoggingField(req, PAYMENT_EXTERNAL_ID, chargeId)
+  
   if (!chargeId) {
     logger.error('ChargeId was not found in request', getLoggingFields(req))
     return false
   } else if (!cookie.getSessionCookieName() || !cookie.isSessionPresent(req)) {
-    logger.info('Session cookie is not present', {
-      ...getLoggingFields(req),
-      referrer: req.get('Referrer'),
-      url: req.originalUrl,
-      method: req.method
-    })
-    return false
-  } else if (!cookie.getSessionVariable(req, 'ch_' + chargeId)) {
+    if(req.originalUrl && req.method.toLowerCase() === 'post' && req.originalUrl.includes(`/card_details/${chargeId}/3ds_required_in`)) {
+      return chargeId
+    } else {
+      logger.info('Session cookie is not present', {
+        ...getLoggingFields(req),
+        referrer: req.get('Referrer'),
+        url: req.originalUrl,
+        method: req.method
+      })
+      return false
+    }
+  } else if (!cookie.getSessionVariable(req, 'ch_' + chargeId)) { 
     logger.info('ChargeId was not found on the session', {
       ...getLoggingFields,
       referrer: req.get('Referrer'),
--- a/test/cypress/integration/card/frontend-state-cookie-removed.spec.js
+++ b/test/cypress/integration/card/frontend-state-cookie-removed.spec.js
@@ -1,0 +1,67 @@
+const cardPaymentStubs = require('../../utils/card-payment-stubs')
+
+describe('Frontend state cookie removed', () => {
+
+  const tokenId = 'be88a908-3b99-4254-9807-c855d53f6b2b'
+  const chargeId = 'ub8de8r5mh4pb49rgm1ismaqfv'
+
+  const validPayment = {
+    cardNumber: '4444333322221111',
+    expiryMonth: '01',
+    expiryYear: '30',
+    name: 'Valid Paying Name',
+    securityCode: '012',
+    addressLine1: '10 Valid Paying Address',
+    city: 'London',
+    postcode: 'E1 8QS',
+    email: 'validpayingemail@example.com'
+  }
+
+  const createPaymentChargeStubsEnglish = cardPaymentStubs.buildCreatePaymentChargeStubs(tokenId, chargeId, 'en')
+  const checkCardDetailsStubs = cardPaymentStubs.checkCardDetailsStubs(chargeId)
+
+  beforeEach(() => {
+    // this test is for the full process, the session should be maintained
+    // as it would for an actual payment flow
+    Cypress.Cookies.preserveOnce('frontend_state')
+  })
+
+  describe('Should show "Your payment session has expired" if cookie is removed before submitting the payment', () => {
+    it('Should setup the payment and load the page', () => {
+      cy.task('setupStubs', createPaymentChargeStubsEnglish)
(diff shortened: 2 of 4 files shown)


test/services/charge_param_retriever_test.js
```diff
@@ -52,7 +68,16 @@ describe('charge param retreiver', function () {
     assert.strictEqual(chargeParam.retrieve(VALID_GET_RESPONSE), 'foo')
   })
 
-  it('should return false if the charge param is in THE BODY and has session', function () {
+  it('should return THE ID  if the charge param is in THE BODY and has session', function () { //?? name of test
```
oswaldquek, line 71: remove `//?? name of test` please

app/services/charge_param_retriever.js
```diff
@@ -7,11 +7,14 @@ const cookie = require('../utils/cookies')
...
 exports.retrieve = req => {
   const chargeId = req.params.chargeId ? req.params.chargeId : req.body.chargeId
-  setLoggingField(req, PAYMENT_EXTERNAL_ID, chargeId)
 
+  setLoggingField(req, PAYMENT_EXTERNAL_ID, chargeId)
+  
   if (!chargeId) {
     logger.error('ChargeId was not found in request', getLoggingFields(req))
     return false
+  } else if (req.originalUrl && req.method.toLowerCase() === 'post' && req.originalUrl.includes(`/card_details/${chargeId}/3ds_required_in`)) {
```
oswaldquek, line 16: probably safer to assert that the `frontend_state` and `frontend_state_2` cookies are not present too.

test/services/charge_param_retriever_test.js
narinderchana: done

app/services/charge_param_retriever.js
narinderchana: done

oswaldquek:
We're going to write some cypress regression tests for these so hold off reviewing this for now.

test/cypress/integration/card/frontend-state-cookie-removed.spec.js
```diff
@@ -0,0 +1,67 @@
+const cardPaymentStubs = require('../../utils/card-payment-stubs')
+
+describe('Frontend state cookie removed', () => {
```
oswaldquek, line 3: happy for this test to be removed if deemed unnecessary

app/services/charge_param_retriever.js
```diff
@@ -7,20 +7,25 @@ const cookie = require('../utils/cookies')
...
   } else if (!cookie.getSessionCookieName() || !cookie.isSessionPresent(req)) {
-    logger.info('Session cookie is not present', {
-      ...getLoggingFields(req),
-      referrer: req.get('Referrer'),
-      url: req.originalUrl,
-      method: req.method
-    })
-    return false
-  } else if (!cookie.getSessionVariable(req, 'ch_' + chargeId)) {
+    if(req.originalUrl && req.method.toLowerCase() === 'post' && req.originalUrl.includes(`/card_details/${chargeId}/3ds_required_in`)) {
```
sfount, line 17: I'm concerned about the increasing complexity and side effects of this method. The pattern isn't established by this pull request but it may make it even harder to reason about in the future. 

The method is called `retrieve` but I think if we tried to encapsulate what it was doing it would be something along the lines of `safelyGetChargeExternalIdFromRequestOrFailIfCookieSessionIsInvalid` (exaggerated).

Adding this logic would make it `safelyGetChargeExternalIdFromRequestOrFailIfCookieSessionIsInvalidIfNotThe3DSRequiredInRoute`.

I wonder if we would be able to split the logic here so that requiring a valid cookie session would be a separate middleware from the one that gets the details of the charge from Connector. 

That then might look something like this from a very high level

```js 
const { requireValidCookieSession, fetchChargeDetails } = require('some/middleware')

// doesn't require a valid cookie session
app.post('/3ds/required/in', fetchChargeDetails, inHandler)

// does require a valid cookie session
app.post('/3ds/required/out', requireValidCookieSession, fetchChargeDetails, outHandler)
```

With this proposal if a particular route should require a session is declared at the top level and the logical flow can be followed. It's trying to avoid having to know the ins and outs of how all of the middleware work to understand which paths will require sessions and which will not (the fact that the middleware `retrieveCharge` will call a method `charge_param_retriever.retrieve` which will work one way for some routes and work another way for others for example).

test/cypress/integration/card/frontend-state-cookie-removed.spec.js
sfount: ➕ I believe this is already covered in `test/controllers/secure_controller_test.js`

test/integration/card_details_errors_ft_tests.js
```diff
@@ -146,6 +146,12 @@ describe('chargeTests', function () {
   })
 
   describe('Enter card details page — errors ', function () {
+    it('should error without frontend state cookie', function (done) {
```
sfount, line 149: Is it possible to have an integration test that visits the 3ds in endpoint specifically, once with a valid state cookie and once without a valid state cookie? That would ensure all the cookie + session logic is working at a high level.

I believe these routes are already covered in `/test/integration/three_d_secure_ft_tests.js` so it may be possible to just add assertions to that suite.

test/services/charge_param_retriever_test.js
```diff
@@ -52,7 +68,16 @@ describe('charge param retreiver', function () {
...
-  it('should return false if the charge param is in THE BODY and has session', function () {
+  it('should return THE ID if the charge param is in THE BODY and has session', function () {
     assert.strictEqual(chargeParam.retrieve(VALID_POST_RESPONSE), 'foo')
   })
+
+  it('should return THE ID if the charge id is in the 3ds post url', function () {
+    assert.strictEqual(chargeParam.retrieve(VALID_POST_RESPONSE_FOR_3DS), 'foo')
+  })
+
+  it('should return THE ID if the charge id is in the EPDQ 3ds post url', function () {
```
sfount, line 79: 👍

sfount (commented):
This code change looks like it will safely do what is required by the ticket without breaking any of the current flows. I've written some thoughts on code complexity introduced by introducing additional side effects to middleware that's already not clear. 

If this should be merged quickly I think this code will work, otherwise I'm happy to discuss how we could simplify this to make the code more robust for any similar future changes.

stevenjmesser:
This is marked as Done in JIRA. But has it been released?

sfount:
I believe this was superseded by:

https://github.com/alphagov/pay-frontend/pull/1786
https://github.com/alphagov/pay-frontend/pull/1790
https://github.com/alphagov/pay-frontend/pull/1794
https://github.com/alphagov/pay-frontend/pull/1798
https://github.com/alphagov/pay-frontend/pull/1799
https://github.com/alphagov/pay-frontend/pull/1802

stevenjmesser:
Ah, should we ask @narinderchana to close the PR in that case?

PR #1782, opened 2020-09-16
PP-7113: Remove non-descriptive labelling

oswaldquek:
We had 2 labels for the `input` month element which causes accessibility issues.


app/views/charge.njk
```diff
@@ -198,7 +198,7 @@
                 <div class="govuk-form-group govuk-clearfix govuk-!-margin-bottom-7 {% if highlightErrorFields.expiryMonth %} error{% endif %} expiry-date" data-validation="expiryMonth">
                   <fieldset class="govuk-fieldset" role="group" aria-describedby="expiry-date-hint">
                     <legend>
-                      <label class="govuk-label govuk-label--s" id="expiry-date-lbl" for="expiry-month">
+                      <div class="govuk-label govuk-label--s" id="expiry-date-lbl" for="expiry-month">
```
iqbalgds, line 201: Just need to update the end tag as well on line 207

iqbalgds (changes requested):

PR #1783, opened 2020-09-16
PP-7116 - Remove text when DDC is sent to Worldpay

nimalank7:
Description:
- Currently when the service has Google/Apple Pay enabled and is using Worldpay 3DS Flex a paying user will see 'Pay with Google/Apple Pay' and 'Enter payment details' with a spinner. This is a bug because when the DDC is sent to Worldpay we hide most but not all of the elements on the card details page. This PR fixes this bug by hiding the elements which contain the above text.
- Since a user cannot click on 'Submit' and 'Pay with Apple/Google Pay' at the same time we can hide the entire container for Google/Apple Pay
- The confirmPaymentDetailsStubs was modified when being passed into 'DDC times out' test so this has been deep cloned so that subsequent tests won't be affected
- A further PR needs to be made conceptualising how we hide elements
- Cypress test indicates that now the 'Enter with Payment Details' and 'Pay with Google Pay' text has been removed:
Screenshot 2020-09-28 at 11 21 56 am

### DIFF
--- a/app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
+++ b/app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
@@ -5,28 +5,19 @@ const toggleWaiting = () => {
   document.getElementById('spinner').classList.toggle('hidden')
   document.getElementById('error-summary').classList.add('hidden')
 
-  var paymentMethodSubmitElement = document.getElementById('apple-pay-payment-method-submit')
-  if (typeof paymentMethodSubmitElement !== 'undefined' && paymentMethodSubmitElement !== null) {
-    paymentMethodSubmitElement.classList.add('hidden')
+  var paymentDetailsHeader = document.querySelector('.govuk-heading-l.web-payment-button-heading')
+  if (typeof paymentDetailsHeader !== 'undefined' && paymentDetailsHeader !== null) {
+    paymentDetailsHeader.style.display = "none"
   }
 
-  paymentMethodSubmitElement = document.getElementById('google-pay-payment-method-submit')
-  if (typeof paymentMethodSubmitElement !== 'undefined' && paymentMethodSubmitElement !== null) {
-    paymentMethodSubmitElement.classList.add('hidden')
+  var applePayContainer = document.querySelector('.apple-pay-container')
+  if (typeof applePayContainer !== 'undefined' && applePayContainer !== null) {
+    applePayContainer.style.display = "none"
   }
 
-  var paymentMethodDivider = document.getElementById('apple-pay-payment-method-divider')
-  if (typeof paymentMethodDivider !== 'undefined' && paymentMethodDivider !== null) {
-    paymentMethodDivider.classList.add('hidden')
-    paymentMethodDivider.classList.remove('pay-divider')
-    document.getElementById('apple-pay-payment-method-divider-word').classList.add('hidden')
-  }
-
-  paymentMethodDivider = document.getElementById('google-pay-payment-method-divider')
-  if (typeof paymentMethodDivider !== 'undefined' && paymentMethodDivider !== null) {
-    paymentMethodDivider.classList.add('hidden')
-    paymentMethodDivider.classList.remove('pay-divider')
-    document.getElementById('google-pay-payment-method-divider-word').classList.add('hidden')
+  var googlePayContainer = document.querySelector('.google-pay-container')
+  if (typeof googlePayContainer !== 'undefined' && googlePayContainer !== null) {
+    googlePayContainer.style.display = "none"
   }
 }
 
--- a/test/cypress/integration/worldpay-3ds-flex-ddc/payment-with-session-id.spec.js
+++ b/test/cypress/integration/worldpay-3ds-flex-ddc/payment-with-session-id.spec.js
@@ -1,4 +1,7 @@
+const { expect } = require('chai')
+const _ = require('lodash')
 const cardPaymentStubs = require('../../utils/card-payment-stubs')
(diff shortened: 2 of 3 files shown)


app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
```diff
@@ -5,28 +5,19 @@ const toggleWaiting = () => {
...
-    paymentMethodDivider.classList.remove('pay-divider')
-    document.getElementById('apple-pay-payment-method-divider-word').classList.add('hidden')
-  }
-
-  paymentMethodDivider = document.getElementById('google-pay-payment-method-divider')
-  if (typeof paymentMethodDivider !== 'undefined' && paymentMethodDivider !== null) {
-    paymentMethodDivider.classList.add('hidden')
-    paymentMethodDivider.classList.remove('pay-divider')
-    document.getElementById('google-pay-payment-method-divider-word').classList.add('hidden')
+  webPaymentContainer = document.querySelector('.google-pay-container')
```
alexbishop1, line 18: I think we should give this a different name rather than reassigning the existing variable.

test/cypress/integration/worldpay-3ds-flex-ddc/payment-with-session-id.spec.js
```diff
@@ -24,10 +32,36 @@ describe('Worldpay 3ds flex card payment flow', () => {
...
+          cardDetails: '1111'
+        },
+        tokenizationData: {
+          type: 'PAYMENT_GATEWAY',
+          token: '{"signature":"MEQCIB54h8T/hWY3864Ufkwo4SF5IjhoMV9hjpJRIsqbAn4LAiBZz1VBZ+aiaduX8MN3dBtzyDOZVstwG/8bqJZDbrhKfQ\\u003d","protocolVersion":"ECv1","signedMessage":"{\\"encryptedMessage\\":\\"I2fg4rbEzgRHpvJqN4pHa7Zh93eGePLCKwHljtTfgWELsLOx0Or1cBQ6giKNUrJza3DqhS0AO+Qoar44J2HBryMRaiuSIi/un0zsNV9cfmiLSHNjHNnATkYrfY4u/Or2uSeuAaxLEt3d91NhHWKqf6BelNme182onG23GvOqd4RAukUW7RJ04eouaCIvBisdt866uq/9B3jJb0QiT91ifZ8C/bfScnBFPL4AX0X3G+B7418wSTtUYrMVBnyhNJS8T2Aw9oW8s7c9pra4PI9cHfcu22opRxzSS9snBF39uTwk8c+Pjj3G8yfNI0biDkHNAUiA1YuBW5CgHeJZ/FhayAsAPzfMmI4qei9nTzIEPlDkkGsqFnamCYIg8N9SM51YV8NGS0MQTIYmJg3GHl2D/We30D6dvYSWfLDDJNATAgb3l+4powoxogb28cwE9vLjFhOF/GChrGRpM845E9r1q0tNfg\\\\u003d\\\\u003d\\",\\"ephemeralPublicKey\\":\\"BC9B7aoVhvPzR54m8P1CkGFDSyuxl04iqu22SvnrlLgZEoH3EvpBNKPyMS/nLIe3mJ+cw26GglqMs00B7EEEs0I\\\\u003d\\",\\"tag\\":\\"lOylJZZF3xdVpsqpl5bKgZdsxRPetMRvcKu9WnmFQ/k\\\\u003d\\"}"}'
+        },
+        type: 'CARD'
+      }
+    },
+    payerEmail: 'name@email.fyi',
```
alexbishop1, line 52: This could be real email address: use .test instead.

test/cypress/integration/worldpay-3ds-flex-ddc/payment-with-session-id.spec.js
```diff
@@ -134,4 +216,15 @@ describe('Worldpay 3ds flex card payment flow', () => {
...
+  setUpAndCheckCardPaymentPageWithGooglePay()
+  describe('Web payment text is not displayed on DDC spinner screen', () => {
+    it('iframe post should still submit the form but without the worldpaySessionId', () => {
+      cy.task('setupStubs', confirmPaymentDetailsStubs)
+      cy.get('#card-details').submit()
+
+      cy.get('.google-pay-container').should('have.attr', 'style', 'display: none;')
+      cy.get('.govuk-heading-l.web-payment-button-heading').should('have.attr', 'style', 'display: none;')
+    })
+  })
```
alexbishop1, line 229: Would it be worth having an Apple Pay equivalent of this?

alexbishop1 (changes requested):
Looks nice overall.

test/cypress/integration/worldpay-3ds-flex-ddc/payment-with-session-id.spec.js
nimalank7: I did think about this but given that the code to hide Apple Pay is identical to Google Pay the test should be sufficient to verify whether worldpay-3ds-flex-ddc.js is functioning as intended. Furthermore given that we are going to reconceptualise the hiding we may refactor this test to simply check whether a `<div class="hidden-stuff>` is hidden/visible so seems redundant to add another test.

test/cypress/integration/worldpay-3ds-flex-ddc/payment-with-session-id.spec.js
nimalank7: Done

app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
nimalank7: Done.

alexbishop1 (approved):

PR #1786, opened 2020-09-17
PP-7118 normalise middleware

narinderchana:
Description:
- Added middlewareStackFor3dsWithSessionValidation in order to  clean up the code in readiness for removing session check from 3ds iframe posts.

### DIFF
--- a/app/routes.js
+++ b/app/routes.js
@@ -34,7 +34,7 @@ exports.bind = function (app) {
   // charges
   const card = paths.card
 
-  const middlewareStack = [
+  const standardMiddlewareStack = [
     csrfCheck,
     csrfTokenGeneration,
     actionName,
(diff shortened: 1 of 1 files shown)


app/routes.js
```diff
@@ -57,18 +61,18 @@ exports.bind = function (app) {
...
+  app.get(card.auth3dsRequiredIn.path, middlewareStackFor3dsWithSessionValidation, threeDS.auth3dsRequiredIn)
+  app.post(card.auth3dsHandler.path, [actionName, middlewareStackFor3dsWithSessionValidation, resolveService, stateEnforcer], threeDS.auth3dsHandler)
 
   // Apple Pay endpoints
   app.post(paths.applePay.session.path, applePayMerchantValidation)
 
   // Generic Web payments endpoint
-  app.post(paths.webPayments.authRequest.path, retrieveCharge, resolveLanguage, webPaymentsMakePayment)
-  app.get(paths.webPayments.handlePaymentResponse.path, retrieveCharge, resolveLanguage, webPaymentsHandlePaymentResponse)
+  app.post(paths.webPayments.authRequest.path, middlewareStackFor3dsWithSessionValidation, webPaymentsMakePayment)
```
alexbishop1, line 74: The naming doesn’t seem right if we’re going to apply this to web payments stuff

alexbishop1 (changes requested):
The idea is fine but I have concerns about the naming

app/routes.js
narinderchana: What do you think about middlewareStackWithSessionValidation, that will mean the next one would be middlewareStackWithoutSessionValidation.

Any suggestions would be good.

alexbishop1: I’m a bit conflicted on this. I’m not sure if we should be naming based on what the middleware does (which runs the risk of getting out of date when we decide to add something new to a middleware stack) or the sorts of pages it applies to (which runs into problems when you decide your 3DS middleware would be ideal for something non-3DS).

Right now we have `middlewareStack`, which is used for most pages, so maybe that could become `standardMiddlewareStack`. And this new one could be `chargeCookieRequiredMiddlewareStack` (because that’s basically all it does). And the one that doesn’t validate the session could be `chargeCookieNotRequiredMiddlewareStack`. Perhaps not the most future-proof but we can easily rename them if our needs change.

narinderchana: I understand that the names can become out date, but are simply trying to make the code more readable. Hopefully when changes are made in the future they will also give the naming the same due diligence.

I will go with your suggestions for now.

alexbishop1 (approved):

PR #1787, opened 2020-09-17
PP-7195 Update Node to 12.18.4

sfount:
Following a security patch, update Node patch version to 12.18.4.

**To review this PR:** 
- [ ] Verify that the SHA hashes match what is available through dockerhub
- [ ] Everything is happy in CI/ test environments


alexbishop1 (approved):
Version numbers are correct and SHAs match what `docker pull node:12.18.4-alpine3.12` and `docker pull node:12.18.4` ouptut (but not what https://hub.docker.com/_/node?tab=tags&name=12.18.4-alpine3.12 and https://hub.docker.com/_/node?tab=tags&name=12.18.4 say for some reason).

PR #1789, opened 2020-09-17
PP-7195 Fix build and run Dockerfile alpine upgrade

sfount:
This Dockerfile should be removed as soon as possible as it introduces
additional flake in the build/ update release process.

- Updated alpine `python` dependency doesn't resolve, specify that we
will be using the `python2` package
- Set the engines version to a paas compatible minimum until we can
patch that separately


danworth (approved):
Looks good, the `build_and_test.Dockerfile` could probably be replaced with a multi-stage docker build (might have a look in the next fire break): https://docs.docker.com/develop/develop-images/multistage-build/

sfount:
@danworth it certainly can! I believe the reasons for having two files are legacy and should definitely be removed (along with all the additional dependancies that are needed)

https://github.com/alphagov/pay-toolbox/blob/master/Dockerfile does this is a reasonable way and CI also uses the actual Dockerfile instead of the CI box
* \+ Docker layer caching
* \+ less places to upgrade

PR #1790, opened 2020-09-17
PP-7118 separate session cookie validation and retrieve charge

narinderchana:
Description:
- charge_param_retriever.js is validating session and returning the charge, the session validation is move to session.js & charge_param_retriever.js removed.
- retrive_charge was validating the session, build charge and view rendering. Now only builds charge and renders view.
- Removing nested arrays for middleware
- In routes we replace retreiveCharge in the middleware with enforceSessionCookie, retrieveCharge


package-lock.json
```diff
@@ -22092,7 +22092,7 @@
         },
         "got": {
           "version": "9.6.0",
-          "resolved": "https://registry.npmjs.org/got/-/got-9.6.0.tgz",
+          "resolved": "http://registry.npmjs.org/got/-/got-9.6.0.tgz",
```
sfount, line 22095: I suspect the package lock change here was accidental, I don't think this PR should do this

app/middleware/csrf.js
```diff
@@ -8,17 +8,17 @@ const logger = require('../utils/logger')(__filename)
 const { getLoggingFields } = require('../utils/logging_fields_helper')
 const session = require('../utils/session')
 const responseRouter = require('../utils/response_router')
-const chargeParam = require('../services/charge_param_retriever')
+const validateSessionCookie = require('../services/validate_session_cookie')
```
sfount, line 11: I wonder if it would make sense for these actions to be grouped by the fact they all perform actions on "sessions" rather than being split across multiple areas as they currently are. 

Naming is always challenging in programming but frontend seems to have a folder called `services` and `utils` and uses the two interchangeably. 

The name `validate` also doesn't imply that a charge ID will be returned.

There already exists a `/utils/session.js`, I think it's quite clear to understand that these are a set of utilities that act on cookie sessions. What do you think of the following:

```js
const {  getChargeIdIfExists } = require('utils/session')

// returns chargeId if exists and a falsey value otherwise
const chargeId = getChargeIdIfExists(req)
```

sfount (commented):
I haven't had much time to look into this but some thoughts on (the particularly confusing legacy) naming.

package-lock.json
narinderchana: fixed

app/middleware/csrf.js
narinderchana: Maybe you are correct that the validation code can belongs in `/utils/session.js.` I think it eventually should return true/false (should only validate).  Then in retrieve_charge will only build the charge, shouldn't check session something like this.

`const chargeId = req.params.chargeId ? req.params.chargeId : req.body.chargeId
req.chargeId = chargeId
 Charge(req.headers[CORRELATION_HEADER]).find(chargeId, getLoggingFields(req))
      .then(data => {
        req.chargeData = data
        setLoggingField(req, GATEWAY_ACCOUNT_ID, data.gateway_account.gateway_account_id)
        setLoggingField(req, GATEWAY_ACCOUNT_TYPE, data.gateway_account.type)
        setLoggingField(req, PROVIDER, data.gateway_account.payment_provider)
        next()
   })
  .catch((err) => {
        responseRouter.systemErrorResponse(req, res, 'Error finding charge in middleware', withAnalyticsError(), err)
   })`

Then 'validate_session_cookie' would eventually look something like

`const {  validateSessionCookie } = require('utils/session')
const validateSessionCookie = validateSessionCookie(req)
if(!validateSessionCookie)){
    responseRouter.response(req, res, 'UNAUTHORISED', withAnalyticsError())
}
next() // not sure this is need
` 
Then in route we will have we have `...,validateSessionCookie, retrieveCharge, ....` or just `retrieveCharge.` let's discuss this at 11:00am

narinderchana:
This is now ready for review, there are no functional changes. Just refactored to move the code to correct files with appropriate test case. Mostly been lift and shift. 

I would like to add an additional test case for routes.js which basically assumes all routes require a valid session cookie, unless explicitly ignored from the test.

 `const routesWithoutSessionCookieValidation = {}`

In this way we force the developer to think about whether route requires session cookie validation.

app/utils/session.js
```diff
@@ -1,7 +1,50 @@
...
 
 // Local dependencies
+const { PAYMENT_EXTERNAL_ID } = require('@govuk-pay/pay-js-commons').logging.keys
+const logger = require('../utils/logger')(__filename)
+const { setLoggingField, getLoggingFields } = require('../utils/logging_fields_helper')
 const cookie = require('../utils/cookies')
 
-exports.createChargeIdSessionKey = chargeId => 'ch_' + chargeId
-exports.retrieve = (req, chargeId) => cookie.getSessionVariable(req, exports.createChargeIdSessionKey(chargeId))
+const createChargeIdSessionKey = function createChargeIdSessionKey( chargeId ){
```
sfount, line 9: Inconsistent spacing and bracket conventions in this file, ideally these things would be caught by a lint. I believe this repo uses 

```suggestion
const createChargeIdSessionKey = function createChargeIdSessionKey(chargeId) {
```

test/middleware/enforce_session_cookie_test.js
```diff
@@ -0,0 +1,79 @@
...
+    })
+
+    it('should call next when frontend_state cookie contains chargeId', done => {
+        const chargeData = validChargeDetails().getPlain()
+        nock(process.env.CONNECTOR_HOST)
+            .get(`/v1/frontend/charges/${chargeId}`)
+            .reply(200, chargeData)
+        enforceSessionCookie(validRequest, response, next)
+
+        const testPromise = new Promise((resolve) => {
```
sfount, line 67: This looks like how this was tested before so potentially we shouldn't prolong this any further by changing it here but this has a number of red flags: 
* Test result is based on an arbitrary timeout instead of a chain of events, this could be flakey if any part CI is slow and could provide false negatives
* What is actually being tested is particularly hard to follow

As this test isn't testing the interaction between the application and the backend that could be stubbed instead of using `nock` OR `next` could be a promise that's passed in and the test depends on the fulfilment of that instead of waiting for a timeout. Happy to chat about that separately.

test/unit/session_test.js
```diff
@@ -0,0 +1,74 @@
...
+    params: {},
+    body: { chargeId: chargeId },
+    method: 'POST',
+    frontend_state: { ch_foo: true },
+    get: () => null
+}
+
+describe('session utils ', () => {
+    describe('createChargeIdSessionKey ', () => {
+        it(`should return chargeId with prefix ch_`, function () {
```
sfount, line 42: _very minor_  but string interpolation (``) doesn't need to be used unless there is dynamic content, in this case favour ('')

sfount (commented):
Looks good to me, the separation of these middleware will result in the middleware stack being much easier to understand and having less side effects or "magic". 

I think the duplicate import in `app/middleware/csrf` needs to be addressed and all of the other comments and minor and could potentially be addressed in follow up PRs

app/middleware/csrf.js
```diff
@@ -8,17 +8,17 @@ const logger = require('../utils/logger')(__filename)
 const { getLoggingFields } = require('../utils/logging_fields_helper')
 const session = require('../utils/session')
 const responseRouter = require('../utils/response_router')
-const chargeParam = require('../services/charge_param_retriever')
+const validateSessionCookie = require('../utils/session')
```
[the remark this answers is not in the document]
narinderchana: done

app/utils/session.js
narinderchana: done

test/unit/session_test.js
narinderchana: done

app/middleware/csrf.js
narinderchana: Implemented this pattern

test/middleware/enforce_session_cookie_test.js
narinderchana: May be can fix this in a separate PR, as you said it was originally test like this. If we are good with that, are you able to approve the change.

We discuss the appropriate test in detail later.

narinderchana:
I also formatted the following code test/middleware/enforce_session_cookie_test.js, as it was not correct.

app/middleware/csrf.js
```diff
@@ -44,7 +43,14 @@ exports.csrfCheck = (req, res, next) => {
   }
 }
 
-function csrfValid (csrfToken, chargeSession, req) {
+function fetchAndValidateChargeId(req) {
+  if (session.validateSessionCookie(req)) {
```
sfount, line 47: 👍 a much clearer change, great work!

sfount (approved):
Change looks good to me 👍 all comments addressed and test standards can be addressed in a follow up as discussed

PR #1794, opened 2020-09-22
PP 7118 remove 3ds in cookie validation

narinderchana:
Description:
- Remove session cookie validation for /card_details/:chargeId/3ds_required_in & /card_details/:chargeId/3ds_required_in requests

### DIFF
--- a/test/integration/three_d_secure_ft_tests.js
+++ b/test/integration/three_d_secure_ft_tests.js
@@ -226,15 +226,17 @@ describe('chargeTests', function () {
 
         nock(process.env.CONNECTOR_HOST)
           .get('/v1/frontend/charges/' + chargeId).reply(200, chargeResponse)
-        const cookieValue = cookie.create(chargeId)
+
         const data = {
           PaRes: 'aPaRes',
           MD: 'aMD'
         }
-        postChargeRequest(app, cookieValue, data, chargeId, false, '/3ds_required_in')
+        postChargeRequest(app, false, data, chargeId, false, '/3ds_required_in')
           .expect(200)
           .expect(function (res) {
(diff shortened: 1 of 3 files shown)


test/integration/three_d_secure_ft_tests.js
```diff
@@ -60,6 +60,8 @@ describe('chargeTests', function () {
         getChargeRequest(app, cookieValue, chargeId, '/3ds_required_out')
           .expect(200)
           .expect(function (res) {
+            expect(res.header['set-cookie'][0].startsWith('frontend_state=')).to.be.true
+            expect(res.header['set-cookie'][0].length).to.be.greaterThan(100)
```
alexbishop1, line 64: I think we decided that it wasn’t worth checking we’re setting the cookie here. We just want to check on the 3DS required in pages that we definitely don’t.

alexbishop1 (changes requested):
Looks good: just one thing about the tests.

test/integration/three_d_secure_ft_tests.js
narinderchana: This check was in a few places, not only lines 63/64. I have removed the other reference too.

alexbishop1 (changes requested):
Looks good.

To assist our future selves, could you edit the commit message to explain why we’re doing this? Something like this: “Users arrive at the 3DS required in page via a cross-site request and modern browsers treat cookies as SameSite=lax by default, which means we won’t sometimes won’t get sent the cookie when it’s a POST request. We’ll still require the cookie to continue further on the journey and actually process the 3DS result.”

alexbishop1 (approved):

PR #1807, opened 2020-09-28
PP-7080 Tidy up expiry date validation tests

alexbishop1:
Tidy up expiry date validation tests:

- Stub the current date to make the tests not rely on the actual current date, using 21 September 2020 (date picked to be far enough from the end of the month that different time zones are not going to cause issues)

- Don’t test multiple things in each test

No changes to anything apart from the tests. The test cases themselves are the equivalent to the old ones but some of the details differ to reflect the fact we’re stubbing the current date.


kbottla (approved):
LGTM!

alexbishop1:
21 September also picked because https://www.youtube.com/watch?v=Gs069dndIYk.

PR #1808, opened 2020-09-28
PP-7080 Implement new expiry date validation algorithm

alexbishop1:
Implement new expiry date validation algorithm, which takes a month and year (which have already had their whitespace stripped):

1. If _month_ is not exactly 1 or 2 digits, both in the range 0 to 9, return an `invalidMonth` error
2. If _month_ is 1 digit and that digit is 0, return an `invalidMonth` error
3. If _month_ is 2 digits and those two digits are not 10, 11 or 12, return an `invalidMonth` error
4. If _year_ is not exactly 2 or 4 digits, all in the range 0 to 9, return an `invalidYear` error
5. If _year_ is 4 digits and the first digit is 0, return an `invalidYear` error
6. If _year_ is 2 digits, prefix it with 20 (for example, 22 becomes 2022)
7. If _year_ is greater than 2099, return an `invalidYear` error
8. If _year_ is less than the current year, return an `inThePast` error
9. If _year_ matches the current year and _month_ is less than the current month, return an `inThePast` error
10. Otherwise, the expiry date is valid

Apart from being stricter than the old algorithm about catching invalid input, the only functional change is that years after 2099 are now considered invalid. Previously they were allowed but then normalised to 2 digits and then, depending on the payment gateway, changed back to 4 digits assuming the 21st century (so 2119 could become 19 and then 2019). Since we shouldn’t have to worry about actual card expiry dates later than 2099 for a number of years yet this new behaviour is sensible but eventually we (or our successors) are going to have to deal with years after 2099.

Also add substantially more tests to check this all works.

### DIFF
--- a/app/utils/charge_validation_fields.js
+++ b/app/utils/charge_validation_fields.js
@@ -73,20 +73,41 @@ module.exports = (Card, chargeOptions = { collect_billing_address: true }) => {
 }
 
 // Validation Functions
-function expiryMonth (expiryMonth, allFields) {
-  let expiryYear = allFields.expiryYear
-  if (!expiryMonth || !expiryYear) return 'message'
 
+function expiryMonth (expiryMonth, allFields) {
   // validations for both expiry month and year are done together in 'expiryMonth'
   // it can't be renamed as it all runs off meta programming further up the stack
-  expiryMonth = expiryMonth - 1 // month is zero indexed
-  if (!(/^\d+$/.test(expiryMonth) && expiryMonth >= 0 && expiryMonth <= 11)) return 'invalidMonth'
-  if (![2, 4].includes(String(allFields.expiryYear).length)) return 'invalidYear'
-  if (`${expiryYear}`.length === 2) expiryYear = '20' + expiryYear
-  const cardDate = new Date(expiryYear, expiryMonth)
-  const currentDate = new Date()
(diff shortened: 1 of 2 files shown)


app/utils/charge_validation_fields.js
```diff
@@ -73,20 +73,41 @@ module.exports = (Card, chargeOptions = { collect_billing_address: true }) => {
...
+
+  const expiryYearInt = expiryYear.length === 2 ? 2000 + parseInt(expiryYear, 10) : parseInt(expiryYear, 10)
+
+  if (expiryYearInt > 2099) {
+    return 'invalidYear'
+  }
+
+  const now = new Date()
+
+  if (expiryYearInt < now.getFullYear()) {
```
nimalank7, line 104: Could this not be combined with the below to become `if (expiryYearInt < now.getFullYear() || (expiryYearInt === now.getFullYear() && expiryMonthInt < now.getMonth() + 1)) {
    return 'inThePast'
  }`

alexbishop1: Yeah, it could, but nested conditionals make my head hurt.

nimalank7: Ha sure.

nimalank7 (approved):
Looks good!

PR #1827, opened 2020-10-07
PP-7114 - Fix expiry year validation bug on blur

nimalank7:
Description:
- When a user blurs the expiry year field without entering anything in a validation error is thrown
- The error arises because expiryYear is a required field and has ‘data-last-of-form-group’. When form-validation.js is run as part of setting up the client-side validation an eventListener is attached to the blur event which calls checkValidationInLine when triggered. When this code is run the left disjunct (lastOfgroup && required) is true and so calls checkValidation despite ‘blank’ being true.
- There are a couple of ways to fix this. One way would be to add a check for emptiness before calling checkValidation(input). Another would be to remove the ‘data-last-of-form-group’ attribute. I have chosen the latter as it seems easiest without having to rework the validation code.
- It’s unclear what the purpose of ‘data-last-of-form-group’ is meant to be. It is only used 2x in our code (i) addressLine2 and (ii) expiryYear.
- There are a few Cypress tests which check certain paths in form-validation.js However there are no unit tests for the eventListener code. Writing this unit test would require mocking out various DOM elements which seems beyond the scope of this ticket. Whilst a test would be desirable to verify the fix I have run it on my local environment to ensure the fix is working as intended. Further work is planned to reconceptualise client-side validation

Screenshot 2020-10-07 at 2 11 59 pm
Screenshot 2020-10-07 at 2 12 03 pm


iqbalgds:
Nice 👍

iqbalgds (approved):

PR #1837, opened 2020-10-12
PP-7192: Disable PAN_ONLY Google transactions

oswaldquek:
We want to see what the payment page looks like when PAN_ONLY transactions are
disabled.

PAN_ONLY transactions are Google Pay transactions made on Chrome (usually on a
desktop).
CRYPTOGRAM_3DS transactions are Google Pay transactions made via the Google Pay
app. These transactions are "tokenized".

This PR is not meant to be deployed to prod and will be reverted promptly.


stephencdaly (approved):
Have received assurances that extra care will go into going into making sure that this change isn't deployed to prod.

PR #1858, opened 2020-10-22
PP-6869 Tell search engines not to index card payment pages

alexbishop1:
Add `<meta name="robots" content="noindex, nofollow">` to all card payment pages so search engines don’t index them.

This is the approach specified by:
https://www.gov.uk/service-manual/technology/get-a-domain-name#telling-search-engines-not-to-index-pages

This does not affect the GOV.UK product pages available on the same host.

### DIFF
--- a/app/assets/iframe/worldpay-3ds-flex-ddc.html
+++ b/app/assets/iframe/worldpay-3ds-flex-ddc.html
@@ -3,6 +3,7 @@
   <head>
     <meta charset="utf-8">
     <title>Device data collection for Worldpay 3DS Flex</title>
+    <meta name="robots" content="noindex, nofollow">
   </head>
   <body>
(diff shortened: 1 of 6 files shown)


oswaldquek (commented):
Do we need the meta tag on worldpay-3ds-flex-ddc, auth_waiting and capture_waiting because they don't extend layout.njk?

alexbishop1:
> Do we need the meta tag on worldpay-3ds-flex-ddc, auth_waiting and capture_waiting because they don't extend layout.njk?

We need it on every page. If a page extends `layout.njk`, it inherits the `meta` element from there. If it doesn’t extend `layout.njk` (or extends `layout.njk` but uses `{% block head %}` to override `layout.njk`’s own `{% block head %}`), we need to add it separately.

oswaldquek (approved):

PR #1886, opened 2020-11-03
PP-7351 Google Pay: Accept, User-Agent, IP address go to connector

alexbishop1:
When sending the request to connector to authorise a Google Pay payment, send the `Accept-For-HTML` header (which is the `Accept` header that was sent when the enter card details page was loaded), the `User-Agent` and the IP address because connector needs to send them to Worldpay for 3D Secure.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "package-lock.json",
     "lines": null
   },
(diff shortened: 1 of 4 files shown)


nimalank7 (approved):
Looks good to me

PR #1930, opened 2020-12-03
PP-7477 - Add JSON fields for Google Pay to be SCA compliance

nimalank7:
Description:
- Some of the fields which are required for SCA compliance are already present in the PaymentRequest object.
- This PR adds the countryCode and assuranceDetailsRequired to the object. The documentation doesn't make it clear whether assuranceDetailsRequired is required for SCA compliance so I've included it in anyways. Since Worldpay haven't specified what fields to send in addition to the 3DS parameters we already send adding these fields into our current Google Pay integration shouldn't break anything as we aren't sending the newly added fields along
- See [here](https://developers.google.com/pay/api/android/guides/resources/sca) for the fields Google Pay want us to add

### DIFF
--- a/app/assets/javascripts/browsered/web-payments/helpers.js
+++ b/app/assets/javascripts/browsered/web-payments/helpers.js
@@ -86,7 +86,8 @@ const getGooglePaymentsConfiguration = () => {
     type: 'CARD',
     parameters: {
       allowedAuthMethods: allowedCardAuthMethods,
-      allowedCardNetworks: allowedCardNetworks
+      allowedCardNetworks: allowedCardNetworks,
+      assuranceDetailsRequired: true
     },
     tokenizationSpecification
   }
@@ -102,6 +103,7 @@ const getGooglePaymentsConfiguration = () => {
     transactionInfo: {
       totalPriceStatus: 'FINAL',
       totalPrice: window.paymentDetails.amount,
+      countryCode: 'GB',
       currencyCode: 'GBP'
     },
     allowedPaymentMethods: [cardPaymentMethod]


app/assets/javascripts/browsered/web-payments/helpers.js
```diff
@@ -86,7 +86,8 @@ const getGooglePaymentsConfiguration = () => {
     type: 'CARD',
     parameters: {
       allowedAuthMethods: allowedCardAuthMethods,
-      allowedCardNetworks: allowedCardNetworks
+      allowedCardNetworks: allowedCardNetworks,
+      assuranceDetailsRequired: true
```
alexbishop1, line 90: Are you sure this is right? I thought this was a response property, not a request property.

alexbishop1 (changes requested):
I note our current merchant name is “GOV.UK Pay”. We might want to consider if we should change this to the service organisation name or something (perhaps not for now).

app/assets/javascripts/browsered/web-payments/helpers.js
nimalank7: Yep if we set assuranceDetailsRequired to true in the request then we get assuranceDetailsRequired in the response. See [here](https://developers.google.com/pay/api/android/reference/request-objects#CardParameters). It's unclear whether this is explicitly required from both the email and the documentation for SCA.

alexbishop1: I suppose as long as we don’t blow up when we receive it in the response (which I assume you’ve checked) then it’s good to request it.

alexbishop1 (approved):

PR #1942, opened 2020-12-11
PP-7082 Frontend Design System asset caching

iqbalgds:
- Add feature flag - NGINX_CACHING_ENABLED
- If flag = true then
  - Add add cache headers to the Design System assets
- Want to see if adding this means we can cache assets from the Design system using NGINX.
- Will be testing this only on the Test env.

### DIFF
--- a/server.js
+++ b/server.js
@@ -137,7 +137,13 @@ function initialisePublic (app) {
   app.use('/javascripts', express.static(path.join(__dirname, '/public/assets/javascripts'), publicCaching))
   app.use('/images', express.static(path.join(__dirname, '/public/images'), publicCaching))
   app.use('/stylesheets', express.static(path.join(__dirname, '/public/assets/stylesheets'), publicCaching))
-  app.use('/', express.static(path.join(__dirname, '/node_modules/govuk-frontend/govuk/')))
+
+  if (process.env.NGINX_CACHING_ENABLED === 'true') {
+    app.use('/', express.static(path.join(__dirname, '/node_modules/govuk-frontend/govuk/'), publicCaching))
+  } else {
+    app.use('/', express.static(path.join(__dirname, '/node_modules/govuk-frontend/govuk/')))
+  }
+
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -137,7 +137,13 @@ function initialisePublic (app) {
   app.use('/javascripts', express.static(path.join(__dirname, '/public/assets/javascripts'), publicCaching))
   app.use('/images', express.static(path.join(__dirname, '/public/images'), publicCaching))
   app.use('/stylesheets', express.static(path.join(__dirname, '/public/assets/stylesheets'), publicCaching))
-  app.use('/', express.static(path.join(__dirname, '/node_modules/govuk-frontend/govuk/')))
+
+  if (process.env.NGINX_CACHING_ENABLED === 'true') {
+    app.use('/', express.static(path.join(__dirname, '/node_modules/govuk-frontend/govuk/'), publicCaching))
```
sfount, line 142: I don't think setting this header requires being behind an NGINX proxy, it should be OK to set this uniformly as with the other resources.

This could be naturally reviewed if and when the card payments page is separated from the product pages.

sfount (approved):
This will set the `max-age` header on design system resources served by express to one year if the flag is set.

If this flag will be around for the long haul there is a list of environment variables and their descriptions in the README that it would be good to update.

iqbalgds:
> This will set the `max-age` header on design system resources served by express to one year if the flag is set.
> 
> If this flag will be around for the long haul there is a list of environment variables and their descriptions in the README that it would be good to update.

The plan is to test on the TEST env and then remove it straight away if everything is OK.
So it will not be around in the long term.

server.js
iqbalgds: The Hufflepuff team will be putting in the NGINX layer so that new requests for static assets will not hit Node - but served from NGINX instead.

Good point about reviewing our caching strategy once we seperate product pages.

PR #1964, opened 2021-01-06
PP-7602 upgrade node to 12.20.1

SandorArpa:
## WHAT
- upgrade Node to 12.20.1 for TLS bug fix (CVE-2020-8265)


sfount (approved):
Looks good 👍 

I wonder why we have a `.travis.yml`, we should remove that if we don't run against Travis.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:12.19.0-alpine3.12@sha256:36bc6f7b733b890f6f0041ff31d2e45798533932576473e05245ef74ed09abd8
+FROM 12.20.1-alpine3.12@sha256:e63dd88799eeccf4f2869963bf79fb2aa7fa24aacf22ef9d6603c0c3ee7f4a07
```
sfount, line 1: I don't know how I missed this, I even tried it locally 🤦 . 

I believe this should read `node:12.20.1-alpine3.12@sha256:e63dd88799eeccf4f2869963bf79fb2aa7fa24aacf22ef9d6603c0c3ee7f4a07`

PR #1971, opened 2021-01-11
[PP-7082] Change the order of no-cache to fix non-cacheable static assets.

dj-maisy:
## What?
This moves the `noCache()` closure further down the list of Middleware, so that static assets are not accidentally marked as "not cacheable". Notably, this should allow static assets from the design system to become cacheable by Nginx.

We already tested this change using [this Docker image](https://hub.docker.com/layers/govukpay/frontend/PP-7082-tests-caching/images/sha256-c53f4385d62a95e654dc14a76afe787c588f4e830c315786f489cb99ed52da5b?context=repo)


dj-maisy:
Another PR now addresses this - closing.

PR #1972, opened 2021-01-11
PP-7082 Move no cache headers after static assets 

iqbalgds:
A no cache policy was set on card payments frontend in the past with
some work done on browser history back/forward behaviour.

The existing middleware stack sets this at the beginning which will apply
to both static routes (served through express middleware) and route
controllers responding to requests.

This commit moves the no cache reset below the definition for serving
static assets, this will apply the no cache header only to the actual
route controllers, continuing to respect the intended behaviour of
forcing browsers to request a new copy of route resources.

This allows upstream services (likely proxies) to set cache headers on
requests and cache commonly used assets.

Co-authored-by: Steven Fountain <steven.fountain@digital.cabinet-office.gov.uk>

### DIFF
--- a/server.js
+++ b/server.js
@@ -74,11 +74,6 @@ function initialiseGlobalMiddleware (app) {
     next()
   })
 
-  app.use(function (req, res, next) {
-    noCache(res)
-    next()
-  })
-
   app.use(bodyParser.json())
   app.use(bodyParser.urlencoded({ extended: true }))
 
@@ -151,6 +146,13 @@ function initialiseRoutes (app) {
   router.bind(app)
 }
 
+function setNoCacheHeadersForRoutes (app) {
+  app.use((req, res, next) => {
+    noCache(res)
+    next()
+  })
+}
+
 function listen () {
   const app = initialise()
   app.listen(PORT || 3000)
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -74,10 +74,12 @@ function initialiseGlobalMiddleware (app) {
     next()
   })
 
-  app.use(function (req, res, next) {
-    noCache(res)
-    next()
-  })
+  if (process.env.NGINX_CACHING_ENABLED !== 'true') {
```
sfount, line 77: I don't understand the meaning behind the `NGINX_CACHING_ENABLED` flag as it stands, I believe there are now 3 uses of it, one ordering this middleware and the other setting a caching cookie on the design system assets. 

Do we think this logic needs to fork on an environment flag? I believe the logic is to move the existing no cache headers to after the static assets but before the page level response is sent to allow caching while preserving the reason the no cache was initially implemented. I think this should be fine to do everywhere/ in all environments. 

It would be useful if the commit message could explain why the middleware is moved and what it means as well as what's happening.

sfount (commented):
Change looks good, I'm not sure about the need for the feature flag and question about commit history

server.js
dj-maisy: Just taken another quick look at this - I'm inclined to agree with Steve as it feels like moving the closure down is a bug fix of sorts, rather than trying to add or change a feature. The "no cache" closure feels like it was intended to be a "catch-all" for any dynamic content that doesn't match any of the other rules above it.

dj-maisy (approved):
Nice and tidy! LGTM! 👍

PR #1996, opened 2021-02-03
PP-7618: Temporarily only enable Apple Pay for our live account

oswaldquek:
This is while we test the new merchant identity certificate works.

Exactement la même que https://github.com/alphagov/pay-frontend/pull/1988


danworth (approved):
Looks sensible to me

PR #2108, opened 2021-03-29
PP-7760 Update the accepted card types image sprite - include the new Visa logo

iqbalgds:
Update the Visa logo as per below:

image

The Visa electron logo does not appear to be used in our code.
Left it in for now - will need to confirm if it can be removed in the future.


kbottla (approved):
LGTM!

PR #2126, opened 2021-04-06
PP-7965 Kebabify file names

stephencdaly:
Rename files to match conventions here https://pay-team-manual.cloudapps.digital/manual/coding-standards/node-js-style-guide.html#file-naming-conventions

Did the rename and updating usages in separate commits in an attempt to preserve the file history in git. When both are done at once, it doesn't recognise all the changes as renames.


app/metrics/graphite-reporter.js
```diff
@@ -2,14 +2,14 @@
...
-const graphiteReporter = new metrics.GraphiteReporter(registry, "frontend", process.env.METRICS_HOST, process.env.METRICS_PORT || 8094)
+const registry = new metrics.Report()
+const graphiteReporter = new metrics.GraphiteReporter(registry, 'frontend', process.env.METRICS_HOST, process.env.METRICS_PORT || 8094)
 graphiteReporter.start(1000)
 
-module.exports.getCounter = function getCounter(metricsPrefix) {
+module.exports.getCounter = function getCounter (metricsPrefix) {
   if (!registry.getMetric(metricsPrefix)) {
-    const counter = new metrics.Counter
+    const counter = new metrics.Counter()
```
danworth, line 11: Didn't know it would even work without the `()`

app/utils/email-tools.js
```diff
@@ -25,7 +25,7 @@ const validEmail = email => {
   const match = reValidEmail.exec(email)
   return {
     'local-part': match !== null ? match[1] : false,
```
danworth, line 27: curious why `local-part` isn't updated to remove the `'` as per `domain` below, is it because of the `-`?

danworth (approved):
Nice tidying and linting 🏅  There is a lot of change here albeit pretty simple. I'm confident its ok because the CI passing, the scripts wouldn't load if they couldn't find the newly named files and I assume the linting changes were suggested by `standard`

app/utils/email-tools.js
stephencdaly: yeh it will be because of that

PR #2130, opened 2021-04-07
PP-7968: Add cloudfront domain to the content security policy

oswaldquek:
Frontend needs to download custom branding images and scss files from our cloudfront.


app/middleware/csp.js
```diff
@@ -19,7 +19,7 @@ const CSP_SELF = ["'self'"]
 const frameAndChildSourceCardDetails = ["'self'", 'https://secure-test.worldpay.com',
   'https://centinelapi.cardinalcommerce.com']
 
-const imgSourceCardDetails = ["'self'", 'https://www.google-analytics.com', 'https://www.gstatic.com']
+const imgSourceCardDetails = ["'self'", 'https://www.google-analytics.com', 'https://www.gstatic.com', 'https://d106cm2l6ezls7.cloudfront.net']
```
sfount, line 22: In case there are more 👀  on this: as discussed this will be useful for building it and proving the end to end but this domain will be provided through through the environment by TF before reaching production

oswaldquek:
This has been pushed to AWS test ECR with `PP-7968-custom-branding` tag

PR #2131, opened 2021-04-07
PP-7965 Remove pact dependency from fixtures

stephencdaly:
This change is copying what was already done in selfservice to allow for simplification of Cypress stub setup.

"Pactify" response bodies in pact tests themselves, rather than in the fixtures, so that the fixtures can be called directly from Cypress tests

Don't "pactify" request bodies in pact tests, as they don't need to be.


danworth (approved):
Agreed, this is simpler 🏅

PR #2133, opened 2021-04-07
PP-7965 Simplify Cypress stubbing

stephencdaly:
Follow changes made to selfservice to simplify stub setup.

Remove the `stubs.js` file, and instead move stub creation into helper functions which can be called directly from the tests.


kbottla (approved):
LGTM!

PR #2248, opened 2021-06-08
PP-8194 Use Charge paymentProvider instead of gateway account payment provider

kbottla:
## WHAT

- We currently use gateway account payment_provider during payment journey where needed. Gateway account is going to support multiple payment providers (when PSP switching is supported), so use payment_provider available on charge


stephencdaly (approved):
Looks good!

SandorArpa (approved):
LGTM

PR #2257, opened 2021-06-10
PP-7864 - Payment Pages - improve accessibility of card icons.

iqbalgds:
The card icons get hidden on the Payment pages when in Windows Hight Contrast Mode:
See screenshot: 
image

This is because the Card icons are a Background CSS sprite image.
Background images get hidden in high contrast mode:

See link - go to this section > Background Images might vanish
https://accessabilly.com/accessibility-issues-concerning-windows-high-contrast-mode/

The recommended approach is to use inline images instead - i.e. use the `` tag.

However, that would be a lot of work:
- Need to rewrite the CSS.
- Need to create separate images for each icon.
- Also need to modify the JS that changes colour of the card icons depending if it is being used.

To help with this situation, we can add transparent pixel with `alt text`.

This means that when images are disabled in Window High contrast mode - the user will still `alt text`.
See screenshot below of a browser with images disabled:
image


mrlumbu (commented):
Seems legit, but would be good to see 
- guidance indicating this is best practice
- a screenshot showing this working (I think this is possible using tools you showed in tech team meeting?)

app/views/charge.njk
```diff
@@ -173,14 +173,12 @@
...
-                  >
-                          <span class="govuk-warning-text__assistive">
-                            {{ currentAllowedCard.brand }}
-                          </span>
+                          class="{{ currentAllowedCard.brand }}"
+                          data-key="{{ currentAllowedCard.brand }}"
+                          data-credit="{{ currentAllowedCard.credit }}"
+                          data-debit="{{ currentAllowedCard.debit }}"
+                        >
+                          <img src="/images/transparent-accessibiliity.gif" alt="{{ currentAllowedCard.brand | replace("-", " ") | title }}" />
```
sfount, line 181: Is there not a `.label` property on card brands that will do this for you? 

If it's driven directly by the [Connector resource](https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/cardtype/model/domain/CardBrandLabelEntity.java) then it should.

iqbalgds: I didn't know that - will take a look, thanks

iqbalgds: I will merge this in and then look at that.
Thanks

iqbalgds: Does not seem to be a `.label` property.
Only:
-Brand
- Debit
- Credit

PR #2264, opened 2021-06-14
PP-7864 Payment Pages - Add fieldset tags

iqbalgds:
- Update the Payment page with `Fieldset` tags as per Design system
  guideline: https://design-system.service.gov.uk/components/fieldset/
- Update the `Billing Address` and `email` fields as follows:
  - Wrap the fields in a `fieldset` tag.
  - As DAC recommends, Add a `fieldset > aria-describedby` attribute which points to the `hint` text.
  - Update the `legend` tag so that it only contains only the heading.
  - Move the `hint` text so that it sits next to the `legend` tag.

See screenshot below: 
image

### DIFF
--- a/app/views/charge.njk
+++ b/app/views/charge.njk
@@ -321,152 +321,161 @@
                   <img src="/images/amex-security-code.png" class="amex-cvc hidden" alt="{{ __p("cardDetails.amexcvcTip") }}"/>
                 </div>
                 {% if collectBillingAddress %}
-                  <div class="govuk-form-group govuk-!-width-three-quarters govuk-!-padding-top-4 govuk-!-margin-top-8 pay-!-border-top {% if highlightErrorFields.addressCountry %} govuk-form-group--error{% endif %}" data-validation="addressCountry">
-                    <legend for="address-country" class="govuk-!-margin-bottom-6">
-                      {% if allowApplePay or allowGooglePay %}
-                        <h3 class="govuk-heading-m govuk-!-margin-bottom-1 web-payment-button-heading">{{ __p("cardDetails.billingAddress") }}</h3>
-                      {% endif %}
-                      <h2 class="govuk-heading-m govuk-!-margin-bottom-1 non-web-payment-button-heading">{{ __p("cardDetails.billingAddress") }}</h2>
+                  <div class="govuk-!-width-three-quarters govuk-!-padding-top-4 govuk-!-margin-top-8 pay-!-border-top">
+                      <fieldset class="govuk-fieldset" aria-describedby="address-hint">
+                        <legend>
+                          {% if allowApplePay or allowGooglePay %}
+                            <h3 class="govuk-heading-m govuk-!-margin-bottom-1 web-payment-button-heading">{{ __p("cardDetails.billingAddress") }}</h3>
+                          {% endif %}
+                          <h2 class="govuk-heading-m govuk-!-margin-bottom-1 non-web-payment-button-heading">{{ __p("cardDetails.billingAddress") }}</h2>
+                        </legend>
 
-                      <span class="govuk-hint govuk-!-margin-bottom-2">{{ __p("cardDetails.billingAddressHint") }}</span>
-                    </legend>
-                    <label id="address-country-lbl" for="address-country" class="govuk-label">
-                      <span
-                    class="address-country-label"
-                    data-label-replace="addressCountry"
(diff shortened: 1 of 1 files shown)


app/views/charge.njk
```diff
@@ -321,152 +321,161 @@
...
-                      {% if allowApplePay or allowGooglePay %}
-                        <h3 class="govuk-heading-m govuk-!-margin-bottom-1 web-payment-button-heading">{{ __p("cardDetails.billingAddress") }}</h3>
-                      {% endif %}
-                      <h2 class="govuk-heading-m govuk-!-margin-bottom-1 non-web-payment-button-heading">{{ __p("cardDetails.billingAddress") }}</h2>
+                  <div class="govuk-!-width-three-quarters govuk-!-padding-top-4 govuk-!-margin-top-8 pay-!-border-top">
+                      <fieldset class="govuk-fieldset" aria-describedby="address-hint">
+                        <legend>
+                          {% if allowApplePay or allowGooglePay %}
+                            <h3 class="govuk-heading-m govuk-!-margin-bottom-1 web-payment-button-heading">{{ __p("cardDetails.billingAddress") }}</h3>
+                          {% endif %}
```
mrlumbu, line 329: could you explain to me what the purpose of this bit is (I know this isn't really a change you've made, just curious)

app/views/charge.njk
```diff
@@ -321,152 +321,161 @@
...
                   or(gatewayAccount.emailCollectionMode === 'OPTIONAL') %}
-                  <div class="email-container govuk-!-width-three-quarters govuk-!-padding-top-4 govuk-!-margin-top-8 pay-!-border-top">
-                    <div class="govuk-form-group {% if highlightErrorFields.email %} govuk-form-group--error{% endif %}" data-validation="email">
-                      <legend for="email" class="govuk-!-margin-bottom-6">
-                        {% if allowApplePay or allowGooglePay %}
-                          <h3 class="govuk-heading-m govuk-!-margin-bottom-1 web-payment-button-heading">{{ __p("cardDetails.contactDetails") }}</h3>
-                        {% endif %}
-                        <h2 class="govuk-heading-m govuk-!-margin-bottom-1 non-web-payment-button-heading">{{ __p("cardDetails.contactDetails") }}</h2>
+                  <div class="govuk-!-width-three-quarters email-container govuk-!-padding-top-4 govuk-!-margin-top-8 pay-!-border-top">
+                    <fieldset class="govuk-fieldset" aria-describedby="email-hint">
```
mrlumbu, line 443: not sure why we have fieldset for email address?

mrlumbu (commented):
Couple of fairly noob questions. 
Happy that you have made the changes described, and I have spun up locally and checked things work as expected

app/views/charge.njk
iqbalgds: The Apple & Google pay code can get confusing.

`{% if allowApplePay or allowGooglePay %}` - This is a flag that states whether the PSP has Google or Apple pay enabled.  This is set in the Node code.

Then the `H3` heading tag is displayed with the text `Billing Address`.
When Apple or Google pay is not present - then this heading is set as a `H2` instead further down the page.

Please let me know if you have any more questions.

app/views/charge.njk
iqbalgds: Good question - as you don’t usually have a `fieldset` for a single field.
We seem to have a `legend` and a hint for the `Email` field but no `fieldset` tag - which is incorrect.

app/views/charge.njk
mrlumbu: I was more wondering why

iqbalgds: The layout of the `Payment Page / Charge page` changes depending on if the `Apple / Google` Pay buttons are visible.

We have an extra **heading**  when the Apple/Google pay buttons are visible.  So all the existing headings will need to drop down a heading level.  I.e. A `h2` will need to become a `h3`.

Please let me know if you have any more questions.

app/views/charge.njk
iqbalgds: Just had another look - I think we should leave the legend tag in for now.
As it links the heading `Contact details` with the field `email`.

Also it seems like you can have a valid `fieldset` with just 1 field.
Please see this website:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset (edited)

mrlumbu (approved):

PR #2265, opened 2021-06-14
Pp 7864 payment page update layout

iqbalgds:
- Update the `Charge` & `Confirm` page layouts based on findings from user research.
- Move the `Payment summary` to the middle and top of the page.
- Charge page:
  - Move the `accepted card types` and card icons above the `card number` input. 
- Make sure the layout works correctly when Apple & Google pay is visible.
  - Rename the classes
    - `web-payment-button-button` -> `web-payment-button-section`
    - `non-web-payment-button-button` -> `non-web-payment-button-section`
    - Show the `payment-summary` in the correct sections depending if the `apple` or `google` pay buttons are visible.

# Charge page

## Old layout:
image

## New layout:
image

# Confirm page

## Old layout
image

## New layout
image

### DIFF
--- a/app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
+++ b/app/assets/javascripts/browsered/worldpay-3ds-flex-ddc.js
@@ -5,7 +5,7 @@ const toggleWaiting = () => {
   document.getElementById('spinner').classList.toggle('hidden')
   document.getElementById('error-summary').classList.add('hidden')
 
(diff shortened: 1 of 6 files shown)


mrlumbu (changes requested):
Mostly looks good. In testing I found that the summary on confirm page has `position: sticky` which makes it go weird when scrolling down (see screenshot). Suggest removing `position: sticky`
Screenshot 2021-06-16 at 09 41 50

mrlumbu (approved):

PR #2268, opened 2021-06-15
Pp 7864 payment page add transparent pixel for Google pay button

iqbalgds:
- Google pay
  - Make it use an inline image of the `Google pay` logo instead of a background image
  - This is better for acessibility for users with High contrast needs.
  - Fix the focus state
    - The button becomes yellow when in focus state.  This would make the `Google Pay`
      image difficult to see.
    - Now the image inverts colours when in `focus` state using CSS.
    - Hover state - make sure the image goes back to normal
- General styling cleanup
  - Make the border radius consistent between the Apple and Google pay buttons

### DIFF
--- a/app/views/charge.njk
+++ b/app/views/charge.njk
@@ -123,7 +123,7 @@
                 <h2 class="govuk-heading-m">{{ __p("cardDetails.webPayments.payWithGooglePay") }}</h2>
 
                 {% set GooglePayButtonHTML %}
-                  <span class="govuk-visually-hidden">Google Pay</span>
+                  <img class="google-pay-image" src="/images/google-pay-logo.svg" alt="Google Pay">
                 {% endset %}
 
                 <form class="govuk-!-width-three-quarters web-payments-container">
--- a/app/assets/sass/modules/_web-payments.scss
+++ b/app/assets/sass/modules/_web-payments.scss
@@ -11,16 +11,16 @@
 
   .google-pay-available & {
     display: block;
-  } 
+  }
 }
 
 .web-payment-button-section{
   display: none;
-  
+
   .apple-pay-available &,
   .google-pay-available & {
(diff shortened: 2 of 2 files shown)


alexbishop1:
Before:
image

After:
image

app/views/charge.njk
```diff
@@ -123,7 +123,7 @@
                 <h2 class="govuk-heading-m">{{ __p("cardDetails.webPayments.payWithGooglePay") }}</h2>
 
                 {% set GooglePayButtonHTML %}
-                  <span class="govuk-visually-hidden">Google Pay</span>
+                  <img src="/images/google-pay-logo.svg" alt="Google Pay" />
```
alexbishop1, line 126: We don’t need the `/` — nobody pretends to do XML any more!

alexbishop1 (changes requested):
The code is more simpler and more straightforward, which is a win.

However, I’m concerned that we might be falling foul of the [Google Pay branding guidelines](https://developers.google.com/pay/api/web/guides/brand-guidelines) by changing the logo from one with [roundrects](https://www.folklore.org/StoryView.py?story=Round_Rects_Are_Everywhere.txt) to harsh, square corners.

iqbalgds:
Fixed the focus state for the Google pay button:

## Normal state:
image

## Focus state:
Image colours get inverted when in focus state so you can see the logo:
image

app/views/charge.njk
iqbalgds: Good spot - removed.

iqbalgds:
Good point about the rounded corners.
I have put them back in.
I have also fixed the `focus` state for the button - so that the images get's inverted when in focus state.

alexbishop1 (approved):
Inverting looks interesting but if it’s what we normally do…

PR #2269, opened 2021-06-15
PP-8248: Update Apple Merchant ID Domain Association file

oswaldquek:
Update the association file for https://www.staging.payments.service.gov.uk/


dj-maisy (approved):
Giving this another go huh? 👍

PR #2295, opened 2021-07-05
PP-8279 Remove consecutive number check from email validation

sfount:
Consecutive numbers are rejected during email validation, this was
initially to prevent card numbers accidentally populating this field.
This has caused some confusion around what a valid email should be -- as
card numbers on their own fail RFC822/ domain checks remove this
explicit check.


kbottla (approved):
LGTM!

PR #2306, opened 2021-07-13
PP-8306 Update Node to 12.22.2

stephencdaly:


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - 12.21.0
+  - 12.22.2
```
kbottla, line 3: Not part of thi PR but don't think we need `travis.yml` anymore

kbottla (approved):

PR #2319, opened 2021-07-21
PP-8256 Use email validation from Notify

stephencdaly:
We were using email validation based on an old spec - RFC 822

This was causing it to be possible for some emails to pass our validation, but then fail Notify's email address validation causing a payment confirmation email not to be sent.

Re-implement Notify's email validation to use instead.

Notify's implementation:
https://github.com/alphagov/notifications-utils/blob/master/notifications_utils/recipients.py#L494-L534 
https://github.com/alphagov/notifications-utils/blob/48c6c822e85d0d1893d2c239e14706cfe0ad8e16/notifications_utils/__init__.py#L14
https://github.com/alphagov/notifications-utils/blob/master/tests/test_recipient_validation.py#L104-L154 

Note: we are now using the `punycode` npm module, which is included in the client-side javascript as it is required for validation the email for apple pay payments. This has changed the size of `application.min.js` from 104KB -> 94KB, probably because the email validation has been separated out so we are no longer requiring "mailcheck" in the client-side javascript, which we weren't using.

Have locked `punycode` at version 1.4.1 as this apparently has broader web browser compatibility.


kbottla (approved):
LGTM!

iqbalgds:
👍🏽 Looks good - like the tests.

iqbalgds (approved):

PR #2321, opened 2021-07-22
PP-8373 Remove payment_provider from gateway_account fixture

kbottla:
## WHAT
- Remove `payment_provider` from gateway account payload in tests. Frontend is using payment_provider on charge everywhere.
- Also set payment_provider available on charge for logging when retrieving charge from connector


SandorArpa (approved):
LOKTM

PR #2387, opened 2021-09-02
PP-8445 Use default billing address country on service

stephencdaly:
If a value is set for the `default_billing_address_country` on the
service, use this as the pre-filled value for the country select.

If no value is set, leave the field blank. This is achieved by adding an
option with the attributes `disabled selected value=""` to the select.
When the select control is converted into the country picker using
javascript from the `govuk-country-and-territory-autocomplete`
npm module, the field will be blank. When javascript is disabled, the
select has the disabled option "Pick a country or territory" selected
by default.

Ended up doing this in the template rather than in normalise-charge.js,
but removed an unused method from normalise-charge.js and improved the
unit tests to make it more understandable and robust.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,3 +1,4 @@
+PORT=3000
 CONNECTOR_HOST=http://localhost:9300
 ADMINUSERS_URL=http://localhost:9700
 CARDID_HOST=http://localhost:9900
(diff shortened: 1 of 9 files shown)


kbottla (approved):
LGTM!

PR #2407, opened 2021-09-17
PP-8575 Update pact dependency

kbottla:
## WHAT
- removes pact@4.3.2 package which is deprecated https://www.npmjs.com/package/pact
- Added new dependency of pact `@pact-foundation/pact`
- and related changes


iqbalgds:
good work 👍🏽

iqbalgds (approved):

PR #2413, opened 2021-09-21
PP-8695: Remove Jenkins ECS deploy step

jfharden:
## WHAT
Jenkins doesn't deploy to ECS now and we have now removed the old EC2 based services so the deploy step is no longer required


katstevens (approved):
Let's see if this works

PR #2425, opened 2021-09-28
PP-7216 update country list

gidsg:
## WHAT
Updates country list based on diff against CSV from https://data.gov.uk/dataset/0b412b57-6934-4f28-b495-640cdc7e8f7f/foreign-commonwealth-and-development-office-geographical-names-index (I wrote a [little python script](https://gist.github.com/gidsg/84e15fb65fe9eea159324369b33f2870) to compare the CSV against our current JSON.)

**Question** should we add the old names as synonyms for the new ones in the autocomplete graph?

## HOW 
Updated country names should appear in country picker on payment page e.g. North Macedonia.
Screenshot 2021-09-28 at 11 18 33

### DIFF
--- a/app/data/countries-autocomplete-graph.json
+++ b/app/data/countries-autocomplete-graph.json
@@ -5482,6 +5482,23 @@
       "en-GB": "Burkina Fasoupper"
     }
   },
+  "nym:Burma": {
+    "edges": {
+      "from": [
+        "country:MM"
+      ]
+    },
+    "meta": {
+      "canonical": false,
+      "canonical-mask": 0,
+      "display-name": false,
+      "stable-name": false
+    },
+    "names": {
+      "cy": false,
+      "en-GB": "Burma"
+    }
+  },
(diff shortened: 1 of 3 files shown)


alexbishop1:
> **Question** should we add the old names as synonyms for the new ones in the autocomplete graph?
Yes, I think that’s a good idea. People may have become used to typing the old names so it would be surprising if they suddenly stopped working.

gidsg:
@alexbishop1 OK I've added a synonym for Swaziland now (the others contain their new name within the old one so will continue to work if people type the old value)

alexbishop1:
Do we also want synonyms for Burma and Macedonia? It seems a little delicate to rely on the fact that they’re substrings of the new names.

gidsg:
> Do we also want synonyms for Burma and Macedonia?

OK I added these now and forced pushed in 1e13ff4

alexbishop1 (approved):

PR #2449, opened 2021-10-11
PP-8716 Web performance - add hash to image assets in Nunjucks file

iqbalgds:
- Add the 'staticify.getVersionedPath' function to app.locals
  - This makes the function available in Nunjuks templates.
  - For production - it will return hashed versions of image assets
    - E.g. Change `image.png` to `image.d5e5c73.png`
- Update Nunjuks templates
  - Update image tags to use `getVersionedPath`
    - E.g. `` to ``

## WHAT
- Add the `staticify.getVersionedPath` to  Express `app.locals`.  So it is available in Nunjucks templates.
Use the Staticify library to do the following:
- For Production - generate Hashed image paths using the `staticify.getVersionedPath`

## HOW 
- Go to the Charge pages and check the inline images - you should see the assets appearing Hashed.
image


iqbalgds:
Closed this - will need to think more about this.

PR #2497, opened 2021-11-08
BAU update product page to 4.9.6

SandorArpa:
## WHAT
- update product page to 4.9.6
- revert (re-generate) `package-lock` to use `lockfileVersion` 1. This happens when npm install is run against an npm version greater than 12, and the package-lock file is also committed with the given branch.


stephencdaly (commented):
Theres 58 additions, 23,456 deletions for the package-lock.json - has something strange happened there?

stephencdaly (approved):
The package lock change reverts a mistaken change made in fd05e1a3c21ff3b75af77213438ec8b1d528f81a that generated it using the wrong node version

PR #2610, opened 2022-01-17
PP-9099 Bump product pages to 4.9.16

iqbalgds:
- Made sure I used the version of Node as specified in Package.json - 12.22.7
      - This means the package-lock.json had to be updated for some minor dependencies
        that are specific for this version of Node.


SandorArpa (approved):
LOKTM

PR #2654, opened 2022-01-31
PP-9148 Cache Cypress with version in key

stephencdaly:
In our workflow for running tests on PRs, we cache Cypress so that it doesn't need to be downloaded and installed for each test run, as this takes a number of minutes. This cache is persisted between action runs.

We were caching with the key `Linux-cypress` not taking into account versioning. This means that if we tried to update our Cypress version, we would probably end up with multiple Cypress versions in the same cache due to restoring the existing version from the cache and then installing the new version when we run `npm install`.

To ensure that we only cache and restore the required version, retrieve the Cypress version number from the package.json file and append this to the cache key.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -35,11 +35,14 @@ jobs:
           path: ~/.npm
           key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
           restore-keys: ${{ runner.os }}-node-
+      - name: Parse Cypress version
+        id: parse-cypress-version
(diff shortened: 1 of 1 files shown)


katstevens (approved):
Yep, looks good to me 👍 Didn't realise `jq` came for free in workflows...

stephencdaly:
> Yep, looks good to me 👍 Didn't realise `jq` came for free in workflows...

Yeh must come in the linux image. I couldn't find anything that said this exactly anywhere but saw that Paas were using it in workflows without having to install it.

PR #2659, opened 2022-02-01
PP-9148 prevent unsafe transitive dependencies being resolved by old Cypress version

nlsteers:
## WHAT
Upgrades:
- Upgrade Cypress to minimum safe version 5

Additions:
- Jenkinsfile now specifies the Cypress version used for integration tests on master builds, see https://github.com/alphagov/pay-jenkins-library/pull/142


app/assets/javascripts/browsered/web-payments/format-card-types.js
```diff
@@ -8,6 +8,8 @@ module.exports = (allowedCardTypes, provider) => {
   const availableNetworks = allowedCardTypes.map(type => {
     if (type.debit || type.credit) {
       return type.brand
+    } else {
+      return []
```
iqbalgds, line 12: I think we may need we change this to return `false` or `null`.
Otherwise it will return true when a empty [] is returned.

package.json
```diff
@@ -127,7 +128,7 @@
     "chai-string": "^1.4.0",
     "cheerio": "^1.0.0-rc.10",
     "chokidar-cli": "latest",
-    "cypress": "3.1.5",
+    "cypress": "5.0.0",
```
iqbalgds, line 131: Would we want a `^` for Cypress?

nlsteers: We now set a specific Cypress version in [the project Jenkinsfile](https://github.com/alphagov/pay-frontend/pull/2659/commits/256241c678e34af2f5a3ec53d6e861bcc179c2c6#diff-e6ffa5dc854b843b3ee3c3c28f8eae2f436c2df2b1ca299cca1fa5982e390cf8) that needs to match the version in the package-lock so the version is pinned.

app/assets/javascripts/browsered/web-payments/format-card-types.js
nlsteers: I'm going to break this into another PR as this one is quite large but I think Array.prototype.map() expects a value to be returned at the end of an arrow function, returning `false` / `null` here would break the conditionals further down the function.

nlsteers: Resolved as no longer part of this PR

stephencdaly (approved):
LGTM!

PR #2668, opened 2022-02-03
PP-9148 Upgrade Standard to v16, replace Sasslint with Stylelint

nlsteers:
## WHAT
Upgrade Standard to version 16, this introduces new rules around usage of `var`. 

Replace Sasslint with Stylelint as it is no longer supported and is including some security vulnerabilities, this adds a new .file for linter rule configuration as well as new rules around CSS property ordering.

Upgrade Cypress to use most recent 5.x.x version.


SandorArpa (approved):
LOKTM

app/assets/javascripts/browsered/web-payments/format-card-types.js
```diff
@@ -5,11 +5,10 @@ module.exports = (allowedCardTypes, provider) => {
     return []
   }
 
-  const availableNetworks = allowedCardTypes.map(type => {
-    if (type.debit || type.credit) {
-      return type.brand
-    }
-  })
+  const availableNetworks = allowedCardTypes.reduce((brands, type) => {
```
iqbalgds, line 8: can we not continue to use `map`?
Just to reduce the number of changes we need to make.

test/cypress/utils/apple-pay-js-api-stubs.js
```diff
@@ -29,9 +29,17 @@ const getMockApplePayClass = (validPaymentRequestResponse, email = null) => {
       this._onvalidatemerchant = value
     }
 
+    get onvalidatemerchant () {
```
iqbalgds, line 32: Are these new functions?

iqbalgds (changes requested):
Again, really good job and amazing effort.

Really interested to hear what others think - might still be worth splitting out into multiple PRs.  So we can revert if we ever need to.

We should also let everyone know - prior to merging in, because people could get confused if they start seeing linting errors.

app/assets/javascripts/browsered/web-payments/format-card-types.js
nlsteers: Standard 16 complains about array transforms that don't return a value in all cases. `.map`, generally, is for transforming data from one array to another without skipping any elements. 

As we are only interested in elements that match our conditional we should be filtering before mapping, we can combine a `.filter` and `.map` into one by using a `.reduce` function here.

test/cypress/utils/apple-pay-js-api-stubs.js
nlsteers: Standard 16 warns against getters without setters. 

We could switch this off I suppose, wdyt?

nlsteers:
This PR has been split into two and moved to a different ticket:

- https://github.com/alphagov/pay-frontend/pull/2749
- https://github.com/alphagov/pay-frontend/pull/2750

Closing.

PR #2692, opened 2022-02-16
PP-9153: Upgrade to Node 16

gidsg:
## WHAT
Upgrade to Node 16 (and associated supporting changes).

Please see individual commit messages for details.

## HOW 
Build steps should pass.


bin/publish-pacts.js
```diff
@@ -27,9 +26,9 @@ async function removePactsNotToBePublished () {
     .forEach(async filename => await unlink(pactDirPath + filename))
 }
 
-async function run() {
-  //make sure pact dir exists, and clean out any files marked to be
-  if (existsSync(pactDirPath)) {
+async function run () {
+  // make sure pact dir exists, and clean out any files marked to be
+  if (await stat(pactDirPath)) {
```
sfount, line 31: Very minor but the interface on these isn't quite the same, stat with throw an `ENOENT` exception very similar to POSIX `stat`, because of this the console output will never be reach, functionally equivalent might be

```
try {
  await stat(pactDirPath)
  await removePactsNotToBePublished()
  await publish()
} catch (e) {
  console.log('Unable to publish pacts', e)
}
```

Or 

```
async function exists(path) {
  try {
    await stat(path)
    return true
  } catch (e) {
    return false
  }
}

async function run() {
  if (await exists(pactDirPath)) //...
}
```

Because this is a one time run synchronous script it just throwing the stat error and ending the process seems like a fine behaviour so don't think this needs to be changed, it might just be confusing to someone later

start.js
```diff
@@ -65,7 +65,7 @@
    * Keep track of processes, and clean up on SIGINT
    */
   function monitor () {
-    fs.writeFileSync(pidFile, process.pid, fileOptions)
+    fs.writeFileSync(pidFile, parseInt(process.pid).toString(), fileOptions)
```
sfount, line 68: Very minor again but I'm not sure why this needs to be run through `parseInt`
```suggestion
    fs.writeFileSync(pidFile, process.pid.toString(), fileOptions)
```

sfount (dismissed):

sfount (approved):

PR #2715, opened 2022-03-01
Multistage

mrlumbu:
Why?
Because fiddling around getting node 16 to work was very painful.
Using a proper multistage build allows us to be completely clear
about versions of things being used, and as a bonus is much faster
as we don't have to repeatedly do things (install, compile)

What?
There are 3 stages defined in the dockerfile, builder, tester,
and final. These can be used by CI to do appropriate things.
Builder installs dependencies and builds public assets.
Tester tests.
Final is the prod ready image, which derives from
a bare node alpine image and then copies only the
stuff it needs from builder, and then prunes dependencies.

Controversies
I have removed the prepublish and prepare tasks in
package.json. Defining these led to them being run as part
of npm install/ci, but this is actually quite annoying.
If we want some catch all task we should define it separately


docker/build_and_test.sh
```diff
@@ -1,25 +0,0 @@
-#!/usr/bin/env bash
```
gidsg, line 1: possibly silly question, if we remove this file, what is doing the equivalent step of running and publishing the PACT files?

mrlumbu: https://github.com/alphagov/pay-jenkins-library/blob/master/vars/buildMultistageAppWithMetrics.groovy#L36-L42

gidsg (approved):
LGTM 👍 🏗️

Dockerfile
```diff
@@ -1,19 +1,52 @@
...
+RUN npm run compile && npm run snyk-protect
+
+
+FROM builder as tester
+WORKDIR /app
+COPY . .
+COPY --from=builder /app/node_modules ./node_modules
+COPY --from=builder /app/govuk_modules ./govuk_modules
+COPY --from=builder /app/public ./public
+RUN npm test
```
sfount, line 26: As discussed it doesn't feel like the docker build process should be concerned with tests like this (especially when its being run as part of a number of different pipelines being phased out and in). I think it would good for this to be called externally

Dockerfile
```diff
@@ -1,19 +1,52 @@
...
 RUN ["apk", "add", "--no-cache", "tini"]
 
-ADD package.json /tmp/package.json
-ADD package-lock.json /tmp/package-lock.json
-RUN cd /tmp && npm ci --production
+WORKDIR /app
+COPY package.json .
+COPY start.js .
+COPY server.js .
+COPY config/ ./config
```
sfount, line 38: Having which folders are required to run the app in the dockerfile feels fairly brittle and tricky to follow (especially if trying to add new things based on current patterns).

Often there is either a single build or configuration with an output folder but this moves it into Docker specific language.

While we generally don't favour copying everything (unpredictable, security, can rely on host machine, etc.) until frontend has a clearer set of production files I think it would be easier to read and less flakey in the future.

A longer term solution to this would be to move everything needed for running the service into a single top level folder (app) and just make sure this is in the final layer.

mrlumbu:
Closing and starting afresh here https://github.com/alphagov/pay-frontend/pull/2720

PR #2719, opened 2022-03-03
PP-9314: add accept-language header to payload

gidsg:
## WHAT
Add `accept-language` header to payload, as this will be used by Connector as part of 3DS flow.

## HOW 
Couldn't find a good place to add a test for this, if you can think of one let me know!

### DIFF
--- a/test/services/normalise-charge.test.js
+++ b/test/services/normalise-charge.test.js
@@ -7,4 +7,37 @@ describe('normalise', function () {
     expect(normalise.expiryDate('01', '17')).to.eql('01/17')
     expect(normalise.expiryDate('01', '2017')).to.eql('01/17')
   })
+
+  it('includes headers in payload', function () {
+    const headers =    {
+      'accept': 'application/json',
+      'x-request-id': 'unique-id',
+      'x-forwarded-for': '127.0.0.1',
+      'user-agent': 'Mozilla/5.0',
+      'accept-language': 'fr;q=0.9, fr-CH;q=1.0, en;q=0.8, de;q=0.7, *;q=0.5'
+    }
+    const card = '4242424242424242'
+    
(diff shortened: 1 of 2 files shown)


alexbishop1 (changes requested):
With regards to tests, it looks like we `charge.ft.test` (and some friends in the same directory), which verifies the actual payload sent to connector. It currently seems to go out of its way to avoid dealing with the `accept_header` and `user_agent_header` so maybe you could update that to test `accept_header`, `user_agent_header` and the new `accept_language_header`.

gidsg:
@alexbishop1 I couldn't figure out how to get those integration tests using the headers, but I added a unit test instead, hope that's sufficient.

test/services/normalise-charge.test.js
```diff
@@ -7,4 +7,37 @@ describe('normalise', function () {
...
+      "addressPostcode": "E1 8QS" },
+      header: headerName => {
+        return headers[headerName]
+      },
+      headers: headers
+    }
+
+    expect(normalise.apiPayload(payload, '4242424242424242').accept_language_header).to.eql('fr;q=0.9, fr-CH;q=1.0, en;q=0.8, de;q=0.7, *;q=0.5');
+    expect(normalise.apiPayload(payload, '4242424242424242').accept_header).to.eql('application/json');
+    expect(normalise.apiPayload(payload, '4242424242424242').user_agent_header).to.eql('Mozilla/5.0');
```
alexbishop1, line 40: Tempting to put the card number in a constant.

alexbishop1 (changes requested):
Unit tests is sufficient for this, I think.

Could you squash the two commits together?

gidsg:
@alexbishop1 OK, have hopefully addressed your latest comments now.

PR #2721, opened 2022-03-03
PP-9153 docker multistage build

mrlumbu:
Change from building frontend using intermediate Dockerfile
to using multistage Dockerfile. This allows docker to do
maximal caching, and makes absolutely sure that versions
(of node, deps etc) are shared in tests and final container.

I have tried to keep the process as close to existing one as
possible, so in building the final image I copy in everything,
then copy in dependencies and built assets from builder.

Other advantages:

only call npm ci once
only do npm run coopile once
no extra confusing docker cruft lying around

Controversies
I have removed the prepublish and prepare tasks in
package.json. Defining these led to them being run as part
of npm install/ci, but this is actually quite annoying.
If we want some catch all task we should define it separately


sfount:
Supersedes https://github.com/alphagov/pay-frontend/pull/2720 and https://github.com/alphagov/pay-frontend/pull/2715

package.json
```diff
@@ -37,7 +37,6 @@
     ]
   },
   "scripts": {
-    "prepare": "npm run compile",
```
sfount, line 40: 🤔 `prepare` is a hook that runs after `npm install` -- is there a chance that this breaks other pipelines/ docs that assume an install will also build assets etc.?

No particular issue with this being removed if not

mrlumbu: Yeah, as I said in https://github.com/alphagov/pay-frontend/pull/2715, I don't think this is a useful thing to do. I'll copy-pasta that reasoning here. Github actions runs that task separately (otherwise tests would fail).

sfount (approved):
Looks good to me, a number of places to improve but they're clearly called out by the change descriptions.

Huge improvement to the image build process no longer relying on a forked docker file.

This journey is covered by lots of tests but may be handy to manually sense check in staging because of the number of moving pieces

PR #2733, opened 2022-03-08
PP-9314: added test for non-present headers

gidsg:
## WHAT
Adds additional unit test.
(I couldn't figure out how to make this work as an integration test but let me know if someone wants to pair on this)

## HOW 
Tests should make sense and pass


alexbishop1 (approved):
I think this is sufficient.

PR #2737, opened 2022-03-09
PP-9375: switch to snyk hosted package

gidsg:
## WHAT
Switch to snyk hosted package

Fixes Deprecation warning
```
⚠ WARNING: Starting from 31 March 2022, the protect command will be removed.
Please use the @snyk/protect package instead.
For more info: https://snyk.co/ueln7
```

## HOW 
build should still pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "cypress:server": "mb | node -r dotenv/config start.js dotenv_config_path=test/cypress/test.env",
     "cypress:test": "cypress run",
     "cypress:test-headed": "cypress open",
-    "snyk-protect": "snyk protect",
+    "snyk-protect": "snyk-protect",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -117,6 +117,7 @@
     "@babel/preset-env": "^7.16.11",
     "@pact-foundation/pact": "^9.17.2",
     "@pact-foundation/pact-node": "^10.17.1",
+    "@snyk/protect": "^1.864.x",
```
alexbishop1, line 120: According to https://www.npmjs.com/package/@snyk/protect the latest version is 1.865 — any reason we’re not using that?

gidsg: @alexbishop1 I just did that so we are not changing versions at the same time as the package source to not conflate things I can update it if you prefer.

alexbishop1 (approved):

PR #2745, opened 2022-03-14
BAU Replace version parsing with shared output

sfount:
Make use of the `outputs` and `needs` links between jobs to share the
Node version parsed from the nvmrc file.


gidsg (dismissed):
🔗 👍

gidsg (approved):

PR #2794, opened 2022-04-06
Pp 9466 send node internal metrics

jfharden:
## WHAT
Add pay-metrics class based on the RFC in pay-decisions: https://github.com/alphagov/pay-frontend/pull/2792

The expectation would be to later extract this class into pay-js-commons for reuse among all node-apps.

One of the goals is to hide that it's statsd under the hood from the user of the metrics class so in future we could replace the metrics backend with something else.

The primary goal though right now is to reintroduce the node garbage collector and event loop stats, and replace the abandoned `metrics` npm module.

## HOW 
For now (while this is draft) the node tests should pass (which include the new tests added against the pay-metrics class.
Also the e2e tests should be relied upon to show the docker build is still ok after the changes to include the required dependencies to build the native modules for event-loop-stats and gc-stats, both of which require c++ native code.


katstevens:
Closing this PR as it's gone stale

PR #2800, opened 2022-04-08
PP 9333 content changes for setting up payment instrument during payment

kundanscorpio:
front end changes to display agreement details on payment screen
agreement description for recurring card payment (TBD
image

### DIFF
--- a/app/views/charge.njk
+++ b/app/views/charge.njk
@@ -177,12 +177,12 @@
                       </p>
                     {% endif %}
                   </label>
-                  
-                  {% if moto and gatewayAccount.motoMaskCardNumberInput %} 
-                    {% set cardNumberInputType = 'password' %}  
+
+                  {% if moto and gatewayAccount.motoMaskCardNumberInput %}
+                    {% set cardNumberInputType = 'password' %}
                     {% set cardNumberAutoCompleteType = 'off' %}
                   {% else %}
-                    {% set cardNumberInputType = 'text' %}   
+                    {% set cardNumberInputType = 'text' %}
                     {% set cardNumberAutoCompleteType = 'cc-number' %}
                   {% endif %}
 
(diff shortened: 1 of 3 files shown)


app/views/charge.njk
```diff
@@ -349,7 +349,12 @@
                     {{ __p("commonConjunctions.or") }}
                   </span>
                   <img src="/images/amex-security-code.png" class="amex-cvc hidden" alt="{{ __p("cardDetails.amexcvcTip") }}"/>
-                </div>
+                 </div>
+                   {% if agreementId %}
```
sfount, line 353: I believe the spec calls for `savePaymentInstrumentToAgreement` flag to be serialised and explicitly checked to show this (placholder) disclaimer. 

For the existing pattern just having an agreement ID might be enough to know this is a payment to set up an agreement but I would be wary of of relying on that logic, especially as this is extended in the future.

To make this explicit this check should be for

```njk
{% if savePaymentInstrumentToAgreement and agreementId %}
```

This will just be for charges consumed by the card frontend (not available through the API, for example)

sfount (commented):
This looks like a great start, I think the story calls for this check to show content to be specifically for the setting up agreement flag on the charge. Happy to chat about what would need to be done to support this

app/views/charge.njk
```diff
@@ -349,7 +349,13 @@
                     {{ __p("commonConjunctions.or") }}
                   </span>
                   <img src="/images/amex-security-code.png" class="amex-cvc hidden" alt="{{ __p("cardDetails.amexcvcTip") }}"/>
-                </div>
+                 </div>
+                   {% if savePaymentInstrumentToAgreement and agreementId  and agreementDescription %}
+                   <div>
```
sfount, line 354: I think the spec calls for this snippet to always be shown by the continue button, closest to the users next action. Here will insert it above collecting billing address (if required by the account) so I think it should move down

sfount, line 354: The other blocks in this form use the design system `govuk-form-group` to help group together components and make sure there's enough space, I think I'd recommend that was done here too:
```html
<div class="govuk-form-group">
    <div class="govuk-inset-text">
        <p class="govuk-body">Your payment details will be saved for:</p>
        <p class="govuk-body">{{ agreementDescription }}</p>
    </div>              
</div>
```

sfount (commented):
Code change and fixtures look good, quick comments on the placement and markup for the snippet

sfount (commented):
It might be good to add a quick Cypress test to assert that this is shown when a payment is made with an agreement available on the fixture and assert that it's not shown when a payment is made without that (current flows). Happy to join on what adding this to the Cypress tests would look like

sfount (approved):

PR #2836, opened 2022-04-27
PP-9333 Cypress coverage for setup agreements disclaimer

sfount:
Include agreements in charge response stub builder.

Assert that normal payment flows will not include the agreements setup
snippet.

Assert that payment flows with the appropriately flagged save payment
instrument to agreement properties will correctly show the agreements
human readable identifier.

### DIFF
--- a/app/views/charge.njk
+++ b/app/views/charge.njk
@@ -533,14 +533,14 @@
             })
           }}
                 {% endif %}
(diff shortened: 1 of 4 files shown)


kundanscorpio (approved):
Changes look good. Tested end to end functionality. works fine

PR #2846, opened 2022-05-04
PP-9289 Country select - update autocomplete attribute

iqbalgds:
- For the `Country` select-  autocomplete attribute
  - Replace it from a random number to `country-disabled-autocomplete`
  - This is so that it is more descriptive.
  - This is a workaround for the Chrome bug which does not respect
    `autocomplete=off`

### DIFF
--- a/app/assets/javascripts/browsered/helpers.js
+++ b/app/assets/javascripts/browsered/helpers.js
@@ -14,8 +14,7 @@ const initialiseAddressCountryAutocomplete = () => {
       displayMenu: 'overlay'
     })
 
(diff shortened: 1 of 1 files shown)


sfount:
Relates to https://github.com/alphagov/pay-frontend/pull/2015

sfount:
Relates to https://github.com/alphagov/pay-frontend/pull/2015

sfount (approved):
Looks like the workaround requiring a unique vale is no longer needed and this meets the need of screen readers better, great work :+1:

I've verified the behaviour locally but it would be worth sense checking in different browsers as it releases through environments. For comparison, the behaviour this was originally addressing is in the linked pull request.

PR #2863, opened 2022-05-16
PP-9613: Bump minimist to 1.2.6

oswaldquek:


package.json
```diff
@@ -73,7 +73,7 @@
     ]
   },
   "overrides": {
-    "minimist": "1.2.5"
+    "minimist": "1.2.6"
```
kbottla, line 76: We have got overrides here https://github.com/alphagov/pay-frontend/blob/master/package.json#L76 so they should be updated too

kbottla (approved):

PR #2868, opened 2022-05-18
PP-9357 Remove 'sass-lint' and add 'stylelint'

iqbalgds:
Use the GDS Stylelint NPM module.
- Add Stylelint dependencies.
- Add Stylelint config - to extend GDS Stylelint config.
- Stylelint issues will be fixed in a seperate ticket that will be prioritised in the future.
- Want to remove `sass-lint` so that we can remove the `minmist` override.

Remove the `minmist` override

Add `eslint-plugin-import` module
- Had to add this NPM module due to an issue with Standard > Eslint and NPM 8.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "start:dev": "nodemon -e js,njk -r dotenv/config start-dev.js",
     "watch": "chokidar app test *.js --initial -c 'npm run test'",
     "lint": "standard --fix",
+    "lint-sass": "stylelint '**/*.scss'",
     "watch-live-reload": "grunt watch",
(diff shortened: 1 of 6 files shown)


stephencdaly (dismissed):

package.json
```diff
@@ -88,6 +86,7 @@
     "compression": "1.7.x",
     "credit-card-type": "6.3.x",
     "csrf": "3.1.x",
+    "eslint-plugin-import": "^2.26.0",
```
kbottla, line 89: Can this be a dev dependency ?

kbottla (approved):

package.json
iqbalgds: Good spot - have updated this.
Just fixing conflicts with package-lock.json

kbottla (approved):

PR #2969, opened 2022-07-15
PP-9847- update Node for security fix

Rkotyank:
## WHAT
Update Node Security fix https://nodejs.org/en/blog/vulnerability/july-2022-security-releases/
This release includes information and patches for:
Three medium severity issues.
Two high severity issues.
High severity issues relate to openSSL (both Node specific and patching openSSL to meet other known vulnerabilities) which shouldn’t be relevant to our web services.
Medium severity issues flagging a number of patches in the request parser in http reportedly opening up for HTTP Request Smuggling (HRS). This is likely relevant to all of our apps and should be considered high priority.
Paired with @katstevens


katstevens:
I've also googled for the Node 16 changelog so we can view what other changes have been introduced since v16.14.2: https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V16.md#16.16.0

nlsteers:
Some bad stuff happened here...

PR #3050, opened 2022-09-13
PP-10059 DVLA Accessibility - stop overwriting labels with error messages

iqbalgds:
# DAC Audit issue identified by DVLA.

### Before changes:
Before error messages: 
image

Error messages - were overwriting labels which is wrong:
image

### After changes

Updated code:
- No longer overwrites labels.
- Error message is a separate element that is no longer embedded in the label element.

image

## Server side validation - charge.njk
- When JS is disabled, make the error messages are not part of the label, make them their own element.

## Credit card hint text:

Credit card hint text is too small - make this bigger:

image

## Full details of changes

- Update Javascript to add error elements rather than overwrite labels.
- When the error is relaced with a new error - create the element again from scratch so that screen readers are alerted.
- Add the error message outside of the label instead of adding it inside the label.
  - This is more semantically correct.
  - Also consistent with Design system.
- Make sure the code works with IE11
  - Use `removeChild` instead of `remove`.
- Credit card hint text
  - Update to use the right size text.
- `charge.njk`
  - For server validation - move error messages so that they are outside of the label element.
  - This is consistent with Design system and semantically correct.
- Update Cypress tests - `moto-mask-payment.cy.test.js` & `payment.cy.test.js`
  - Update selector for error messages.
  - Update tests when entering a valid value - to make sure that there are no errors.

### DIFF
--- a/app/assets/javascripts/browsered/form-validation.js
+++ b/app/assets/javascripts/browsered/form-validation.js
@@ -237,14 +237,20 @@ var init = function () {
   var replaceLabel = function (validation, formGroup) {
     var label = formGroup.querySelectorAll('[data-label-replace]')[0]
     if (label.length === 0) return
+
+    var  errorElement = label.parentNode.parentNode.querySelector('.govuk-error-message') 
+
     if (validation) {
-      label.textContent = validation
-      label.setAttribute('role', 'alert')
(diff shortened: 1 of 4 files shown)


SandorArpa (approved):
LOKTM

PR #3106, opened 2022-10-11
PP-10077 add condition for ebent listener

SandorArpa:
## WHAT
- check source when handling event message

with @alexbishop1


alexbishop1 (approved):
I paired with Sándor on this and am confident it does not require any further review per our security policy.

PR #3111, opened 2022-10-12
PP-10077 Make Google Pay and Apple Pay spinner cover whole page

alexbishop1:
Make it so that when we are processing a Google Pay or Apple Pay payment we hide the majority of the page while showing the spinner (currently none of the page is hidden). In addition, disable all the submit buttons when the spinner is shown (currently only the Google Pay or Apple Pay submit button is disabled).

This brings the appearance of the spinner into line with when we do Worldpay 3DS Flex device data collection. Also make it so that when the spinner is shown for DDC we disable all the submit buttons (currently only the card details submit button is disabled).

with @SandorArpa

### DIFF
--- a/app/assets/javascripts/browsered/helpers.js
+++ b/app/assets/javascripts/browsered/helpers.js
@@ -22,7 +22,73 @@ const initialiseAddressCountryAutocomplete = () => {
   document.getElementsByTagName('head')[0].appendChild(autocompleteScript)
 }
 
+const toggleButton = (button) => {
+  if (button) {
+    button[button.getAttribute('disabled') ? 'removeAttribute' : 'setAttribute']('disabled', 'disabled')
+  }
+}
+
+const toggleSubmitButtons = () => {
+  toggleButton(document.getElementById('submit-card-details'))
+  toggleButton(document.getElementById('apple-pay-payment-method-submit'))
+  toggleButton(document.getElementById('google-pay-payment-method-submit'))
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/browsered/helpers.js
```diff
@@ -22,7 +22,73 @@ const initialiseAddressCountryAutocomplete = () => {
   document.getElementsByTagName('head')[0].appendChild(autocompleteScript)
 }
 
+const toggleButton = (button) => {
+  if (button) {
+    button[button.getAttribute('disabled') ? 'removeAttribute' : 'setAttribute']('disabled', 'disabled')
```
iqbalgds, line 27: Just wondering what the square brackets do here?

iqbalgds (commented):
Looks good - just left one comment

app/assets/javascripts/browsered/helpers.js
alexbishop1: It makes it into either `button['removeAttribute']('disabled')` or `button['setAttribute']('disabled', 'disabled'`, which apparently works (this is copied from existing code).

iqbalgds: Ah - makes sense now.  Clever stuff.

iqbalgds (approved):
Looks great - as discussed, would be good to test on the test env before prod.

PR #3113, opened 2022-10-13
PP-10077 Recover from unexpected Google Pay Ajax response

alexbishop1:
If the Google Pay Ajax call results in an unexpected HTTP status code, hide the spinner and display the main content (which is the same thing we’d do for other errors).

with @SandorArpa and @kbottla


SandorArpa (approved):

PR #3115, opened 2022-10-13
PP-10077 Add Google Pay Worldpay 3DS Flex DDC iframe

alexbishop1:
If both Google Pay and Worldpay 3DS Flex are enabled, add an additional device data collection iframe to the enter card details page. This iframe is identical to the other one except that it has a different ID and will be used exclusively for device data collection for Google Pay payments.

with @SandorArpa


SandorArpa (approved):

PR #3116, opened 2022-10-13
PP-10077 Google Pay: send Worldpay 3DS Flex DDC result to connector

alexbishop1:
When processing a request to `/web-payments-auth-request/google/CHARGE_ID` (sent as an Ajax request from the browser), process the Worldpay 3DS Flex device data collection result and status.

Send the DDC result to `/v1/frontend/charges/CHARGE_ID/wallets/google on connector` as `worldpay_3ds_flex_ddc_result` inside the existing `payment_info` object.

Log the DDC status in the same way we do for a regular card payment for which we performed device data collection.


SandorArpa (approved):
LGTM

kbottla (approved):

PR #3121, opened 2022-10-14
PP-10077 Google Pay: Do not modify PaymentResponse

kbottla:
## WHAT
- We are currently modifying the `PaymentResponse` object returned by GooglePay for the Worldpay DDC result. Stop modifying the response object directly (which looks like it is immutable based on testing) and assign it to a new variable to append DDC result/status.


SandorArpa (dismissed):
LVGTM

SandorArpa (approved):
LGTM

PR #3127, opened 2022-10-17
PP-10077 Nest Google Pay PaymentResponse object in Ajax request

alexbishop1:
When we make the Ajax request to `/web-payments-auth-request/google/CHARGE_ID`, we currently convert the `PaymentResponse` object received from Google Pay to JSON and then add the two new `worldpay3dsFlexDdcResult` and `worldpay3dsFlexDdcStatus` properties to it and then send that as the payload.

Instead, send a JSON payload containing a top-level `paymentResponse` property containing the unmodified `PaymentResponse` object from Google Pay alongside the `worldpay3dsFlexDdcResult` and `worldpay3dsFlexDdcStatus` properties, also at the top-level:

```json
{
  "paymentResponse": … the PaymentResponse object from Google Pay …,
  "worldpay3dsFlexDdcResult": "the DDC result string",
  "worldpay3dsFlexDdcStatus": "the DDC status string"
}
```

This means we are not modifying the `PaymentResponse` object (which is not really ours to modify).

Since Google Pay is currently disabled globally, it’s safe to change both the code that creates the Ajax request and the server- side code that processes the received Ajax request simultaneously without having to worry about supporting both shapes of data simultaneously for a while.

with @SandorArpa

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -140,56 +140,56 @@
         "filename": "test/controllers/web-payments/google-pay/normalise-google-pay-payload.test.js",
         "hashed_secret": "13cba698ab03d9438b0107161da540794080a05d",
         "is_verified": false,
-        "line_number": 24
+        "line_number": 25
       },
       {
         "type": "Base64 High Entropy String",
         "filename": "test/controllers/web-payments/google-pay/normalise-google-pay-payload.test.js",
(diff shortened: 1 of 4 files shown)


SandorArpa (approved):
LGTM

PR #3129, opened 2022-10-17
PP-10077 Tweak selected payload properties logging for Google Pay

alexbishop1:
Change how we log selected properties from the Google Pay payload so that it once again more closely mirrors the actual payload we now receive.


SandorArpa (approved):
LGTM

SandorArpa (approved):
LOKTM

PR #3135, opened 2022-10-20
PP-10077 upgrade cypress to 6.4

SandorArpa:
## WHAT
- upgrading cypress due to functionality that only exist in v6 for testing 'fetch'

with @alexbishop1


alexbishop1 (approved):
I paired with Sándor on this and it’s totes legit.

PR #3141, opened 2022-10-21
PP-10077 Update Google Pay Cypress tests for device data collection

SandorArpa:
Update Google Pay Cypress tests to reflect the fact that we now do device data collection when Worldpay 3DS Flex is enabled (which is the standard configuration now).
    
with @alexbishop1


alexbishop1 (approved):
I paired on this and I’m confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay.

PR #3302, opened 2023-02-02
bau: Disable Apple Pay

oswaldquek:
Apple Pay has to be disabled for now as the payment processing certificate was incorrectly replaced


app/views/charge.njk
```diff
@@ -3,7 +3,7 @@
 {% from "govuk/components/button/macro.njk" import govukButton %}
 
 {% block pageTitle %}
-  {% if allowApplePay or allowGooglePay %}
```
nlsteers, line 6: could we not just set this value to false in the charge controller?

PR #3320, opened 2023-02-15
PP-10675:replaced all usages of set-output

Rkotyank:
## WHAT
Located and replaced all usages of set-output commands in our workflows

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -55,7 +55,7 @@ jobs:
           restore-keys: ${{ runner.os }}-node-
       - name: Parse Cypress version
         id: parse-cypress-version
-        run: echo "CYPRESS_VERSION=$(jq -r '.devDependencies.cypress' package.json" >> $GITHUB_STATE
+        run: echo "CYPRESS_VERSION=$(jq -r '.devDependencies.cypress' package.json" >> $GITHUB_OUTPUT
       - name: Cache Cypress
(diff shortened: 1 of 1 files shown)


.github/workflows/run-tests.yml
```diff
@@ -54,7 +55,7 @@ jobs:
           restore-keys: ${{ runner.os }}-node-
       - name: Parse Cypress version
         id: parse-cypress-version
-        run: echo "::set-output name=CYPRESS_VERSION::$(jq -r '.devDependencies.cypress' package.json)"
+        run: echo "CYPRESS_VERSION=$(jq -r '.devDependencies.cypress' package.json" >> $GITHUB_OUTPUT
```
nlsteers, line 58: easy to miss but you've left off a `)` here

nlsteers:
The `actions/cache` version we're using in these workflows is giving the same warnings. 

Looking at the version we use (`136d96b4aee02b1f0de3ba493b1d47135042d9c0`) there's a newer version (`6998d139ddd3e68c71e9e398d8e40b71a2f39812`) available that might resolve these.

katstevens:
Interesting - looks like we skipped that version! https://github.com/alphagov/pay-frontend/pull/3244

katstevens (approved):

katstevens:
I've told Dependabot to un-ignore v3.2.x of the action - it's opened a shiny new PR which I'll deal with separately 😸 https://github.com/alphagov/pay-frontend/pull/3321

PR #3368, opened 2023-03-17
PP-10839 Update `frontend` csp re google pay changes

nlsteers:
## WHAT

- google recently updated guidance for payment handler api usage, we now need to ensure the domains of HTTP requests sent from the browser are added to the connect-src directive
- see this blog post for more information: https://developer.chrome.com/blog/payment-handler-csp-connect-src


jfharden (approved):

sfount (approved):
It would be good for this to be associated with a ticket (one around the `.card` work would be fine) but also good to get this out -- maybe the PR title can be updated

PR #3457, opened 2023-05-16
PP-10947 Remove unused `eslint` NPM modules

iqbalgds:
- Remove `eslint` & `eslint-plugin-import` NPM modules.
- We do not use Eslint directly - we use Standard to carry out our JS linting.


SandorArpa (approved):
LOKTM

PR #3458, opened 2023-05-16
PP-10947 - Remove unused linting NPM modules

iqbalgds:
- Remove `lint-staged`
  - Believe this was overlooked when husky / pre-commit hooks were on Frontend.
  - No being used - so can be removed
- Remove `stylelint-order`
  - Not being used in the `stylelint` config file.
  - Can be removed.
  - We use the GDS global stylelint config instead.


SandorArpa (approved):
LGTM

PR #3461, opened 2023-05-16
PP-10947 Remove the `moment-timezone` module - use core JS functions instead

iqbalgds:
- We were using `moment-timezone` NPM module to give us the example expiry year hint
  - Which is the current year + 2
- Remove the module and just use core JS functions instead.

### DIFF
--- a/app/services/example-card-expiry-date.js
+++ b/app/services/example-card-expiry-date.js
@@ -1,8 +1,6 @@
 'use strict'
 
-const moment = require('moment-timezone')
-
-const getFutureYearAs2Digits = () => moment(new Date()).tz('Europe/London').add(2, 'years').format('YY')
+const getFutureYearAs2Digits = () => ((new Date()).getFullYear() + 2).toString().slice(-2)
 
 module.exports = {
   getFutureYearAs2Digits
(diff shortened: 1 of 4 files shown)


SandorArpa (approved):
LOKTM

alexbishop1 (changes requested):
Removing the tests is a bit drastic!

We could put the new logic inside `example-card-expiry-date.js` and update `example-card-expiry-date.test.js` so that it works without `moment-timezone`.

Adding two years to the current date and formatted it to two digits isn’t conceptually complicated but it would be easy for a mistake to slip in so it would be good to have tests to make sure it’s doing the right thing, particularly as it appears on such a high-profile page.

iqbalgds:
> Removing the tests is a bit drastic!
> 
> We could put the new logic inside `example-card-expiry-date.js` and update `example-card-expiry-date.test.js` so that it works without `moment-timezone`.
> 
> Adding two years to the current date and formatted it to two digits isn’t conceptually complicated but it would be easy for a mistake to slip in so it would be good to have tests to make sure it’s doing the right thing, particularly as it appears on such a high-profile page.

Good point - I will add the test back in.  Thanks

iqbalgds:
Added the unit test back.
Not using moment for the test - switched to JS date.
I don't need to set the timezone when setting the date for the purposes of the test - so removed that from the test.

alexbishop1 (approved):
Good stuff! Since UK time matches UTC around the new year, any time zone differences are not going to be important.

PR #3522, opened 2023-06-19
PP-10904 Rename Cypress test for the confirmation page and make Cypress tests standalone

iqbalgds:
- Rename test file from `awaiting-auth` to `confirm` - so that it is more descriptive.


test/cypress/integration/card/confirm.cy.test.js
```diff
@@ -52,9 +48,9 @@ describe('Awaiting auth', () => {
     cy.get('#address-city').type(validPayment.city)
     cy.get('#address-postcode').type(validPayment.postcode)
     cy.get('#email').type(validPayment.email)
-  })
 
-  it('Submitting confirmation should redirect to auth waiting page', () => {
```
kbottla, line 57: I think this test is checking the intermediary auth waiting page which redirects to confirm after a wait. There are separate tests for the confirmation page https://github.com/alphagov/pay-frontend/blob/master/test/cypress/integration/card/payment.cy.test.js#LL150C1-L150C1 

Are there any other tests for auth waiting page?

iqbalgds: I see - this makes sense now.
These are the only tests for the auth waiting page.

I now kept the filename the same.

Please take a look.

test/cypress/integration/card/awaiting-auth.test.js
kbottla, line 1: This file name doesn't have `.cy` in the filename. Is this intentional?

Also, the commit title can change as these are not being renamed anymore as initially intended.

iqbalgds: Good spot.

I will add the `cy` back into the filename.

Also, I had updated the commit but Github PR did not get updated so will manually update it.

test/cypress/integration/card/awaiting-auth.test.js
```diff
@@ -21,21 +21,17 @@ const validPayment = {
...
 
 describe('Awaiting auth', () => {
-  beforeEach(() => {
-    // this test is for the full process, the session should be maintained
-    // as it would for an actual payment flow
-    Cypress.Cookies.preserveOnce('frontend_state')
-  })
-
-  it('Should setup the payment and load the page', () => {
+  it('should load the payment page and when card details are submitted then it should go to the awaiting auth page and then getredirect to the confirm page', () => {
```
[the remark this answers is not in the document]
kbottla: don't see updates. Did you push changes?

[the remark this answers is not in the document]
iqbalgds: Sorry about that - just pushed the changes now.

kbottla (approved):

PR #3526, opened 2023-06-19
PP-10904 Make enforce views to state Cypress tests standalone

iqbalgds:


SandorArpa (approved):
LOKTM

PR #3532, opened 2023-06-20
PP-10904 Make email payment cancel Cypress tests standalone

iqbalgds:


test/cypress/integration/card/payment-cancellation.cy.test.js
```diff
@@ -6,24 +6,20 @@ const createPaymentChargeStubs = cardPaymentStubs.buildCreatePaymentChargeStubs(
...
     cy.window().its('chargeId').should('eq', `${chargeId}`)
-  })
 
-  it('Show show the cancel page', () => {
+    cy.task('clearStubs')
     cy.task('setupStubs', createCancelChargeStub)
 
     cy.get('#cancel').submit()
+
+    cy.log('Show show the cancel page')
```
kbottla, line 21: Is it should show?
```suggestion
    cy.log('Should show the cancel page')
```

PR #3552, opened 2023-07-04
PP-11244 Remove handling for old GET token response

stephencdaly:
Connector has now been updated so that the GET token response uses the FrontendChargeResponse schema for the embedded charge rather than ChargeEntity.

Remove the handling for the old response.

Update `secure.controller.test.js` for this change, and refactor it slightly to make it easier to understand.

Update the Pact test for the GET token endpoint to validate the embedded charge in the response using the new schema.

### DIFF
--- a/app/controllers/secure.controller.js
+++ b/app/controllers/secure.controller.js
@@ -27,24 +27,13 @@ exports.new = async function (req, res) {
   const correlationId = req.headers[CORRELATION_HEADER] || ''
   try {
     const tokenResponse = await Charge(correlationId).findByToken(chargeTokenId, getLoggingFields(req))
-    let chargeId, gatewayAccountId, gatewayAccountType
 
-    // Temporarily support both the new and old response schemas from connector
-    if (tokenResponse.charge.externalId) {
-      chargeId = tokenResponse.charge.externalId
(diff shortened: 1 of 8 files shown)


app/controllers/secure.controller.js
```diff
@@ -27,24 +27,14 @@ exports.new = async function (req, res) {
...
-    if (tokenResponse.charge.externalId) {
-      chargeId = tokenResponse.charge.externalId
-      gatewayAccountId = tokenResponse.charge.gatewayAccount.gateway_account_id
-      gatewayAccountType = tokenResponse.charge.gatewayAccount.type
-    } else {
-      chargeId = tokenResponse.charge.charge_id
-      gatewayAccountId = tokenResponse.charge.gateway_account.gateway_account_id
-      gatewayAccountType = tokenResponse.charge.gateway_account.type
-    }
+    console.log('TOKEN RESPONSE: ' + JSON.stringify(tokenResponse))
```
kbottla, line 30: console.log ?

stephencdaly: Thanks, removed

kbottla (approved):

PR #3571, opened 2023-07-14
PP-10904 Add detect secrets baseline file - before updating to Cypress 12

iqbalgds:


test/.secrets.baseline
```diff
@@ -0,0 +1,116 @@
+{
+  "version": "1.4.0",
```
katstevens, line 2: Do we need this additional file if we've already got a `.secrets.baseline` in the root folder?

iqbalgds: Good spot - just removed that file.
We only need the one in the root folder.
Thanks

katstevens (approved):

PR #3575, opened 2023-07-14
PP-10904 Make payment Cypress test standalone.

iqbalgds:
- In previous PR https://github.com/alphagov/pay-frontend/pull/3533, the payment Cypress test was updated so that it was not dependent on the `preserveOnce` method.
- However, forgot to remove the actual `preserveOnce` method.
- This PR removes this method - which is required before upgrading to Cypress 12.


SandorArpa (approved):
LOKTM

PR #3601, opened 2023-08-07
bau: Remove PaaS manifest file

oswaldquek:


katstevens (approved):
Good spot - I thought we'd got rid of all of these

PR #3613, opened 2023-08-15
PP-11323 hide wallet options for recurring payments

SandorArpa:
## WHAT

We currently only offer wallet payments for one-off payments. Although it is possible to use recurring payments (payments taken when setting up agreements), it needs additional backend integration which is not in place for both Stripe/Worldpay.

So we should hide wallet options even if allow_apple_pay / allow_google_pay is true for the gateway account

- do not render wallet payments if the payment is a recurring one (savePaymentInstrumentToAgreement is present)
- add Cypress tests

### DIFF
--- a/app/views/charge.njk
+++ b/app/views/charge.njk
@@ -3,7 +3,7 @@
 {% from "govuk/components/button/macro.njk" import govukButton %}
 
 {% block pageTitle %}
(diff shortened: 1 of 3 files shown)


james-peacock-gds (approved):
LGTM

PR #3626, opened 2023-08-18
PP-11322 Use provider specific env vars wallets

stephencdaly:
Use payment provider-specific environment variables to enable/disable wallet payments.

For both Worldpay and Stripe accounts, if a gateway account ID is in the PAY_TEST_GATEWAY_ACCOUNTS array, ignore the global environment variable so that we are able to test wallet payments for our test gateway accounts only.


SandorArpa (approved):
LGTM

PR #3632, opened 2023-08-23
PP-11364 aend additional data for apple pay

SandorArpa:
## WHAT

Stripe requires us to send some additional data from PKPaymentToken when we create a token for Apple Pay. PKPaymentToken.transactionIdentifier
PKPaymentToken.paymentMethod.displayName
PKPaymentToken.paymentMethod.network

- Send additional fields for both Stripe and Worldpay, as they are available for both.
- Connector should deserialise fields but doesn’t need to do anything with them yet
- Update Pact tests to include additional fields


stephencdaly (approved):
LGTM!

PR #3635, opened 2023-08-24
PP-11365 Send additional `payment_data` to connector for Apple Pay payments

kbottla:
## WHAT

- Currently, we send Apple Pay token.paymentData as JSON to the connector in the field `encrypted_payment_data`. For Stripe payments, we will need to send `paymentData` as is to Stripe. To avoid any issues, when deserialising encrypted_payment_data and then serialising data, we can send paymentData as a String and let connector deserialise payload if required.
- Added a new field payment_data to send token.paymentData as String. Connector doesn't handle the new field yet


stephencdaly (approved):
LGTM!

PR #3646, opened 2023-08-30
PP-11365 Remove `encrypted_payment_data` field

kbottla:
## WHAT
- Removed `encrypted_payment_data` field from Apple Pay payload as Connector no longer requires it


SandorArpa (approved):
LOKTM

PR #3648, opened 2023-08-30
[WIP] PP-11377 Bump Standard to v17

katstevens:
## WHAT
The jump from v16 to v17 isn't huge - the automated fixes below are all for use of shorthand property names. This is not supported in IE, so I've added `eslint` ignores to the user-facing files for these.

I've also removed the dependabot restriction on v17 of standard.


katstevens:
Not sure what's gone wrong with this secrets baseline file - rebasing hasn't helped. Also I think until we drop IE support it may not be worth upgrading to v17 just yet.

PR #3651, opened 2023-08-31
PP-11360 node 18 baseline

nlsteers:
## WHAT

Upgrade `frontend` to Node 18.

Pact tests were failing in this upgrade due to a change in the way localhost is evaluated under Node 18, `/etc/hosts` defines two possible values for localhost; `::1` and `127.0.0.1` (a.k.a IPV6 or IPV4). 

By replacing `localhost` with `127.0.0.1` we explicitly request IPV4 localhost and avoid ambiguity. Preventing unintended requests to `::1` that were throwing ECONNREFUSED errors.


jfharden:
Don't you need to update the Dockerfiles too?

jfharden (approved):
Approving on the understanding of waiting for the e2e tests to pass, it's the only real test of the built container in the tests of the repo

PR #3680, opened 2023-09-19
PP-11462 fix erroneous apple pay option

nlsteers:
## WHAT

In some circumstances, when a user encountered an error during the Google Pay wallet journey they were returned to the card details page and could see Apple Pay as a payment option.

This is because a client-side script modifies the DOM to show/hide wallet options. By wrapping all options in a parent div and applying conditional styling higher up the document tree, we no longer accidentally toggle the visibility of Apple Pay-specific DOM elements.

Additionally, `npm run start:dev` now runs `nodemon` with the `--inspect` flag enabled by default. There's a new launch configuration (vscode only) that can be used to attach to the debugger.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -37,6 +37,14 @@
           ],
           "console": "integratedTerminal",
           "internalConsoleOptions": "neverOpen"
+    },
+    {
+        "type": "node",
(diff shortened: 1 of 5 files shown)


iqbalgds (approved):
Nice work

PR #3684, opened 2023-09-20
PP-11115 fix frontend layout bug

nlsteers:
## WHAT

The continue button was not correctly spaced if a service was configured to collect billing addresses but not email addresses. Conditionally apply margin overrides based on `shouldShowEmail`.

Before:
before

After:
after

After (all fields enabled):
after all fields


SandorArpa (approved):
LGTM

PR #3685, opened 2023-09-20
PP-11463 Fix Google pay description dialog bug

iqbalgds:
- When the payment reference is long and the Google pay dialog appears, then it does not show the `add` button.  Which the user needs to add contact information.
- This is a bug with Google chrome - it does not wrap the payment description text. So its pushes the button too wide so it is not visible.
- This PR truncates the payment description if it is too long when setting up the Google payment.


stephencdaly (approved):
LGTM!

PR #3689, opened 2023-09-21
PP-11470 show auth rejected page

SandorArpa:
## WHAT

We would expect to see an error page when payment was declined after clicking “Pay” in the Google Pay dialog, instead we see a technical problems page.

- refactor auth response controller to handle 400 return code
- add a relevant test

### DIFF
--- a/test/controllers/web-payments/handle-auth-response.controller.test.js
+++ b/test/controllers/web-payments/handle-auth-response.controller.test.js
@@ -128,6 +128,7 @@ describe('The web payments handle auth response controller', () => {
     expect(res.redirect.calledWith(303, '/card_details/chargeId/auth_waiting')).to.be.ok // eslint-disable-line
     done()
   })
+
   it('show capture failed page and delete connector response if connector response is in the session and and capture failed', done => {
     const mockCharge = () => {
       return {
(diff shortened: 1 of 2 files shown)


test/controllers/web-payments/handle-auth-response.controller.test.js
```diff
@@ -274,23 +278,13 @@ describe('The web payments handle auth response controller', () => {
...
-        path: '/handle-payment-response/3/error',
-        amount: '4.99',
-        testingVariant: 'original'
-      }
-    }
+
     requireHandleAuthResponseController(mockCharge, mockNormaliseCharge, mockCookies)(req, res)
     expect(mockCookies.deleteSessionVariable.calledWith(req, `ch_${chargeId}.webPaymentAuthResponse`)).to.be.ok // eslint-disable-line
-    expect(res.render.calledWith('errors/system-error', systemErrorObj)).to.be.true // eslint-disable-line
+    expect(res.render.calledWith('errors/incorrect-state/auth-failure')).to.be.true // eslint-disable-line
```
stephencdaly, line 284: would it be worth checking on the options passed when render is called, as this test was previously? In particular, we'd want to ensure the `returnUrl` is there as expected because the view needs this.

SandorArpa: 👍

stephencdaly (approved):
LGTM! Just make sure to pin this on staging to test on test-12

PR #3694, opened 2023-09-26
PP-11470 temporarily disable some pact test

SandorArpa:
## WHAT

connector code has changed, and now returns 402 on authorisation error. Temporarily disable tests so connector code can be merged.


stephencdaly (approved):
This will temporarily disable these pact tests - we don't have any work in progress that will clash with doing this.

PR #3703, opened 2023-09-29
PP-11395 frontend stripe google pay impl

nlsteers:
## WHAT

Implements Google Pay for Stripe accounts:
- `payment_provider` and `stripePublishableKey` are now available in the dom window, we use them in client-side scripts to:
  - prevent device data collection from taking place for Stripe Google payments.
  - enable a Stripe-specific `tokenizationSpecification` structure when interfacing with Google
- the charge controller determines which Stripe publishable key to use based on the gateway account type
- Stripe-specific `connector` request bodies are handled by `normalise-google-pay-payload.js`
- the `connector` client now understands which provider endpoint to call based on the wallet type of the charge
- new pact testing Stripe-specific Google wallet authorisations

### DIFF
--- a/test/fixtures/wallet-payment.fixtures.js
+++ b/test/fixtures/wallet-payment.fixtures.js
@@ -1,7 +1,7 @@
-const successfulLastDigitsCardNumber = 4242
+const successfulLastDigitsCardNumber = '4242'
 
 const fixtures = {
-  googleAuthRequestDetails: (ops = {}) => {
+  worldpayGoogleAuthRequestDetails: (ops = {}) => {
     const data = {
       payment_info: {
         last_digits_card_number: ops.lastDigitsCardNumber !== undefined ? ops.lastDigitsCardNumber : successfulLastDigitsCardNumber,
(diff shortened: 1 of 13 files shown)


test/fixtures/wallet-payment.fixtures.js
```diff
@@ -21,6 +21,21 @@ const fixtures = {
...
     return data
   },
+  stripeGoogleAuthRequestDetails: (ops = {}) => {
+    return {
+      payment_info: {
+        last_digits_card_number: ops.lastDigitsCardNumber !== undefined ? ops.lastDigitsCardNumber : successfulLastDigitsCardNumber,
+        brand: 'visa',
+        cardholder_name: 'Name',
+        email: 'name@example.com',
+        worldpay_3ds_flex_ddc_result: null,
```
iqbalgds, line 31: As its stripe - do we need to define this?

iqbalgds (approved):
Just left a minor comment.
Approved - if you did want to fix that, can do it as another PR.

test/fixtures/wallet-payment.fixtures.js
nlsteers: we don't, it'll default to null on connectors end if it's not specifically defined here anyway so can be removed.

PR #3706, opened 2023-10-03
PP-11496 Add more logging for wallet payments on the server side

iqbalgds:
- Add logging for errors with the capture.


iqbalgds:
This is no longer required - superceded by more recent changes.

PR #3707, opened 2023-10-03
PP-11037 Accessibility bug - add `aria-hidden=true` to header and footer images

iqbalgds:
- This should have been done when upgrading to design system v4.
- We consider the header and footer images to be decorative. Therefore adding `aria-hidden=true` so that they are ignored by screen readers.


SandorArpa (approved):
LOKTM

PR #3711, opened 2023-10-04
PP-11582 Fix Google Pay repeated error message bug

james-peacock-gds:
Context: Bug - if a Google Pay payment fails multiple times, the same error message is shown multiple times
- Clear existing error messages before displaying the new error message

### DIFF
--- a/app/assets/javascripts/browsered/web-payments/helpers.js
+++ b/app/assets/javascripts/browsered/web-payments/helpers.js
@@ -10,6 +10,7 @@ const showErrorSummary = (title, body) => {
   errorSummary.classList.remove('hidden')
   errorSummary.querySelectorAll('h2')[0].innerText = title
   if (body) {
+    errorSummary.querySelectorAll('ul')[0].innerHTML = ''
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/browsered/web-payments/helpers.js
```diff
@@ -10,6 +10,7 @@ const showErrorSummary = (title, body) => {
   errorSummary.classList.remove('hidden')
   errorSummary.querySelectorAll('h2')[0].innerText = title
   if (body) {
+    errorSummary.querySelectorAll('ul')[0].innerHTML = ''
```
stephencdaly, line 13: Maybe we should clear the contents of the `ul` even if there is no body. I imagine if we called this method with just a title, we would want just that title to be displayed and not the body of the previous error.

james-peacock-gds: Good point. I've now made that change.

stephencdaly (approved):
LGTM!

PR #3718, opened 2023-10-09
PP-11496 Wallet payments - update error handling and logging of the `handle auth response` controller

iqbalgds:
- During a wallet payment - when calling connector, improve the error handling and logging
  - Add a catch all to pick up any unknown status codes returned from connector.

### DIFF
--- a/app/controllers/web-payments/handle-auth-response.controller.js
+++ b/app/controllers/web-payments/handle-auth-response.controller.js
@@ -39,6 +39,7 @@ const handleAuthResponse = (req, res, charge) => response => {
         logger.info('Requesting 3DS1 for Google payment, redirect to auth 3ds page', getLoggingFields(req))
         return redirect(res).toAuth3dsRequired(req.chargeId)
       } else {
+        logger.info('Requesting capture for digital wallet payment', getLoggingFields(req))
         Charge(req.headers[CORRELATION_HEADER])
           .capture(req.chargeId, getLoggingFields(req))
           .then(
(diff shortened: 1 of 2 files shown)


app/controllers/web-payments/handle-auth-response.controller.js
```diff
@@ -70,17 +71,14 @@ const handleAuthResponse = (req, res, charge) => response => {
...
-    case 402:
-    case 500:
+    default:
       logging.failedChargePost(response.statusCode, getLoggingFields(req))
       responseRouter.systemErrorResponse(req, res, 'Wallet authorisation error response', withAnalytics(
         charge,
         { returnUrl: routeFor('return', charge.id) },
         webPaymentsRouteFor('handlePaymentResponse', charge.id))
       )
       break
```
SandorArpa, line 81: No break needed for `default`

SandorArpa (changes requested):

app/controllers/web-payments/handle-auth-response.controller.js
iqbalgds: Thanks - just updated it.

SandorArpa (approved):
LGTM

test/controllers/web-payments/handle-auth-response.controller.test.js
```diff
@@ -297,6 +297,39 @@ describe('The web payments handle auth response controller', () => {
...
+        analyticsId: 'test-1234',
+        type: 'test',
+        paymentProvider: 'sandbox',
+        path: '/handle-payment-response/3/error',
+        amount: '4.99',
+        testingVariant: 'original'
+      }
+    }
+    requireHandleAuthResponseController(mockCharge, mockNormaliseCharge, mockCookies)(req, res)
+    expect(mockCookies.deleteSessionVariable.calledWith(req, `ch_${chargeId}.webPaymentAuthResponse`)).to.be.ok // eslint-disable-line
```
stephencdaly, line 328: For future reference, `sinon.assert.calledWith` is better as it gives a helpful error message about whether it was called or not and if it was called but the args didn't match, it will give a comparison of the expected and actual values.

iqbalgds: Thanks - I will merge this in and raise another PR to update that.

PR #3722, opened 2023-10-10
PP-11496 Update `handle-auth-response` controller tests

iqbalgds:
- Update the test that was added previously to use sinon.asserts.calledWith instead of expect.  As it gives better error messages with what is expected vs. actual.


james-peacock-gds (approved):
LGTM

PR #3746, opened 2023-10-25
PP-11641 Point to generic Google Pay endpoint

james-peacock-gds:
Context: connector's google/worlpay and google/stripe endpoints are being merged into a single endpoint for Google Pay authorisations

- point frontend wallet authorisation requests to the new wallets/google endpoint
- point consumer tests to new connector provider states that include stubbing for Worldpay and Stripe
- remove the decline and error consumer tests, to be reintroduced when sandbox authorisations are enabled.

### DIFF
--- a/app/controllers/web-payments/payment-auth-request.controller.js
+++ b/app/controllers/web-payments/payment-auth-request.controller.js
@@ -37,7 +37,6 @@ module.exports = (req, res, next) => {
   const chargeOptions = {
     chargeId,
     wallet,
-    paymentProvider,
     payload
   }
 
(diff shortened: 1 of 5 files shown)


stephencdaly (approved):
Looks good to me! I'm sure you're already intending to do the following, but just make sure to:
1. Pin frontend on staging and then test Google pay on the test environment for both Worldpay and Stripe before allowing it to be deployed to production. Make sure that it is calling the connnector endpoint we expect in both cases.
2. Add an acceptance criteria for the frontend story for Google Pay sandbox to add back in the Pact test for a declined payment.

PR #3749, opened 2023-10-26
PP-11638 Add endpoint for client-side logging

stephencdaly:
Add a `/log/:chargeId` POST route. This can be called by the client-side JavaScript to do logging.

This accepts a POST request with a body containing a `code`, which maps to a message to be logged. We want to do this rather than accepting any string to be logged to avoid log injection.

This endpoint always returns a 200, even if we do not log due to the `code` being invalid or missing. This is because we do not want to interrupt client-side JavaScript running if this fails.

In the client-side JavaScript call this endpoint to log steps in the wallet payment journeys, such as when the button is clicked to start an Apple Pay or Google Pay payment, and also to log when certain errors are caught. When we call the loggng endpoint, we do not handle the Promise resolving/rejecting as we don't need to handle the response in any way.

### DIFF
--- a/app/assets/javascripts/browsered/web-payments/apple-pay.js
+++ b/app/assets/javascripts/browsered/web-payments/apple-pay.js
@@ -1,7 +1,12 @@
 'use strict'
 
 const { prepareAppleRequestObject, showErrorSummary } = require('./helpers')
-const { toggleSubmitButtons, showSpinnerAndHideMainContent, hideSpinnerAndShowMainContent } = require('../helpers')
+const {
+  toggleSubmitButtons,
+  showSpinnerAndHideMainContent,
+  hideSpinnerAndShowMainContent,
+  sendLogMessage
+} = require('../helpers')
 const { validateEmail } = require('../../../../utils/email-validation')
 const { email_collection_mode } = window.Charge || {} // eslint-disable-line camelcase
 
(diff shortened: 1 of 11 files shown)


app/assets/javascripts/browsered/web-payments/apple-pay.js
```diff
@@ -26,7 +31,8 @@ module.exports = () => {
         })
       } else {
         ga('send', 'event', 'Apple Pay', 'Error', 'Merchant ID not valid')
-        return session.abort()
+        return sendLogMessage(window.chargeId, 'ApplePayMerchantIdNotValid')
```
iqbalgds, line 34: I was wondering if it is worth writing this this as follows? 
As it I think it means that it might be a bit quicker - as its not waiting for `finally` block to execute?

```suggestion
        sendLogMessage(window.chargeId, 'ApplePayMerchantIdNotValid')
        return session.abort()
```

iqbalgds (commented):
Really good work 👍
I think the `fetch` looks good.
Just added one comment.

app/assets/javascripts/browsered/web-payments/apple-pay.js
stephencdaly: Yeh, maybe don't need to wait for the promise to resolve, as we're doing in the other places

stephencdaly: Have updated

iqbalgds (approved):
Great work - required a lot of code changes, thanks

PR #3755, opened 2023-11-06
PP-11638 Move client-side logging for wallet available

stephencdaly:
We were logging that wallet payment methods were shown to the end user fewer times than they were being selected.

Move where we log whether the the wallet methods are available to the same place we add the classes to make the buttons visible. What was probably happening was a race condition between the JavaScript that shows the button and the JavaScript that checks whether the button is shown to log the message.

Remove the code that just sends some Google Analytics events as we do not use Google Analytics currently, and have not for some time.

Additionally, log when we remove the Wallet buttons because our scripts haven't loaded properly.


iqbalgds:
Looks good - some Cypress test's are failing

stephencdaly:
> Looks good - some Cypress test's are failing

Thanks - I think what I tried isn't going to work. Will have to try something else

PR #3758, opened 2023-11-06
PP-11638 Move client-side logging for wallet available

stephencdaly:
We were logging that Google Pay is available fewer times than Google Pay was being selected - which obviously shouldn't be possible.

We think this was happening because we were doing the logging in a DOMContentLoaded event listener, based on the presence of the 'google-pay-available' class on the document body.

The 'google-pay-available' class is added in a callback when the Promise to check whether Google Pay payments are available has resolved. It must be possible for this promise to resolve after the code in the DOMContentLoaded event listener that checks for the presence of the class has executed.

To avoid this, add another DOMContentLoaded event listener to the inline script in `charge.njk`, after the Promise to check whether Google Pay is available has resolved, which will log that Google Pay is available. We need to do this in a DOMContentLoaded event listener to ensure the script used to do the logging has loaded.

Move the logging for Apple Pay being available to `charge.njk` to mirror the logging for Google Pay, although the check for Apple Pay isn't asynchronous so we think the previous logging was probably working correctly.

Also, remove the code that sends the Google Analytics events as we do not use Google Analytics currently, and have not for some time.

### DIFF
--- a/app/assets/javascripts/browsered/web-payments/index.js
+++ b/app/assets/javascripts/browsered/web-payments/index.js
@@ -10,20 +10,6 @@ const { sendLogMessage } = require('../helpers')
 const paymentMethodForms = Array.prototype.slice.call(document.getElementsByClassName('web-payments-container'))
 const standardMethodForm = document.getElementById('card-details')
 
-const initApplePayIfAvailable = () => {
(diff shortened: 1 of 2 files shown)


iqbalgds (approved):
Great work

PR #3759, opened 2023-11-07
PP-11638 Add additional client-side logging

stephencdaly:
Add additional client-side logging in places where we were previously sending events to Google Analytics.

For most of these, we could probably get the same information from the server-side logs. However, logging the outcome of authorisation requests in the client-side logging will make it easier to quickly track the outcome of wallet payments and build dashboards to represent the user journey.


james-peacock-gds (approved):
LGTM!

PR #3761, opened 2023-11-09
PP-11638 Add additional client-side logging

stephencdaly:
Add additional client-side logging in places where we were previously sending events to Google Analytics.

This is just logging the response the client receives from the server so we can identify whether there was an error making the request to the server to authorise the payment. If the request was made successfully, regardless of the outcome (authorisation success, authorisation rejected or authorisation error), the server will return a 200 status code to the client. In this case log that the response has been successfully processed - but the client does not know the outcome of the authorisation request at this point, but we can look at the server logs to find that out.


SandorArpa (approved):
LGTM

PR #3762, opened 2023-11-09
PP-11578 Support Google Pay for sandbox

stephencdaly:
- Show Google Pay option for sandbox accounts, if the `allow_google_pay` flag is set to true for the gateway account
- At this time, do not support an environment variable for disabling Google Pay on sandbox. We suspect we will not have the need to disable it globally, as if it stops working on Sandbox it will not affect paying users.

- Add pact test for making a request to connector to authorise a sandbox Google Pay payment when connector returns a success response. This commit does not add a Pact test for a declined response - this will come in a later commit as we need a special state for this.


james-peacock-gds (approved):
LGTM!

PR #3763, opened 2023-11-09
PP-11714 CSP reporting endpoint

nlsteers:
## WHAT

- added a new endpoint to handle csp violation reports
- updated the `cardDetailsCSP` and `worldpayIframeCSP` helmet configurations to send reports to new endpoint
- added new paramterised middleware functions to handle csp reporting and domain exclusions
  - hardening measures:
    - payload size limit
    - rate limiting
    - payload shape validation
- tests and helper functions

### DIFF
--- a/app/middleware/csp.js
+++ b/app/middleware/csp.js
@@ -1,38 +1,40 @@
 const helmet = require('helmet')
+const Sentry = require('../utils/sentry').initialiseSentry()
+const paths = require('../paths')
+const express = require('express')
+const { rateLimit } = require('express-rate-limit')
+const logger = require('../utils/logger')(__filename)
+const hasSubstr = require('../utils/has-substr')
 
 const sendCspHeader = process.env.CSP_SEND_HEADER === 'true'
 const enforceCsp = process.env.CSP_ENFORCE === 'true'
-const cspReportUri = process.env.CSP_REPORT_URI
-const environment = process.env.ENVIRONMENT
 const allowUnsafeEvalScripts = process.env.CSP_ALLOW_UNSAFE_EVAL_SCRIPTS === 'true'
-
-const sentryCspReportUri = `${cspReportUri}&sentry_environment=${environment}`
 
 // Script responsible for setting 'js-enabled' class, extends GOV.UK frontend `layout` which we have no control over
 // and never changes
-const govUkFrontendLayoutJsEnabledScriptHash = "'sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU='"
+const govUkFrontendLayoutJsEnabledScriptHash = '\'sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU=\''
 
-const CSP_NONE = ["'none'"]
-const CSP_SELF = ["'self'"]
+const CSP_NONE = ['\'none\'']
+const CSP_SELF = ['\'self\'']
 
 // Worldpay 3ds flex iframe - frame and child must be kept in sync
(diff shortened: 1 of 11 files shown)


app/middleware/csp.js
```diff
@@ -1,56 +1,58 @@
...
+const logger = require('../utils/logger')(__filename)
+const hasSubstr = require('../utils/has-substr')
 
 const sendCspHeader = process.env.CSP_SEND_HEADER === 'true'
 const enforceCsp = process.env.CSP_ENFORCE === 'true'
-const cspReportUri = process.env.CSP_REPORT_URI
-const environment = process.env.ENVIRONMENT
 const allowUnsafeEvalScripts = process.env.CSP_ALLOW_UNSAFE_EVAL_SCRIPTS === 'true'
 
-const sentryCspReportUri = `${cspReportUri}&sentry_environment=${environment}`
```
katstevens, line 9: Do we no longer need to specify the environment to Sentry? I can't quite see where we're doing it now

nlsteers: we send via the DSN now using the Sentry Node client

test/middleware/csp.test.js
```diff
@@ -1,6 +1,147 @@
...
+        .set('x-forwarded-for', '9.9.9.9')
+        .send(validPayload)
+        .expect(204)
+    }
+
+    await request(underTest)
+      .post('/test')
+      .set('x-forwarded-for', '9.9.9.9')
+      .send(validPayload)
+      .expect(429)
```
katstevens, line 142: As discussed, let's have a test for the level 2 `trust proxy`  (given that's what we'll be using).

katstevens (commented):
I'm satisfied the combination of rate limiting and request validation should cover potential issues, given we're not logging any values from the request.

A Node person should give a 2nd pair of eyes on the code itself!

server.js
```diff
@@ -184,6 +204,7 @@ function initialise () {
   const app = unconfiguredApp
   if (NODE_ENV !== 'test') {
     app.use(metrics.initialise())
+    app.use(paths.csp.path, cspMiddlewareStack) // CSP violation monitoring
```
stephencdaly, line 207: As this is setting up a particular endpoint that executes a middleware stack, rather than us binding middleware to a bunch of endpoints, I think this would be easier to understand and find if we moved this to `routes.js` and used `app.post` to set up the endpoint.

stephencdaly (commented):
Solution looks really good. Just left a suggestion about moving where we set up the `/csp-report` endpoint.

Also [report-uri ](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri) is now deprecated, so as discussed we should move over to using `report-to`. Although this can be done separately, as our existing implementation would be affected by this anyway.

nlsteers:
> Also [report-uri ](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri) is now deprecated, so as discussed we should move over to using `report-to`. Although this can be done separately, as our existing implementation would be affected by this anyway.

@stephencdaly after a brief look there's bit more to `report-to` than just a straight substitution with `report-uri`, I'll create a polcat story to look into it as there's some behaviour re iframes that will probably be relevant to us.

server.js
nlsteers: good suggestion, have amended

test/middleware/csp.test.js
nlsteers: yep, the tests did not adequately validate this behaviour, have amended

stephencdaly (approved):
LGTM!

PR #3769, opened 2023-11-14
PP-11468 Update Apple Pay Cypress tests to test the Node endpoint

iqbalgds:
with @SandorArpa

Apple Pay Cypress tests
- The Apple Pay Cypress tests was not testing the Node endpoint to authorise and capture a charge for wallet payments.
- Instead, the Cypress test was mocking out the entire JS `fetch` call.
- So the test was only covering the client side code.
- This PR removes the mocked `fetch`.
- The test will now pass through all calls to the Node endpoint.
- A similar thing was done to the Google Pay Cypress tests: https://github.com/alphagov/pay-frontend/pull/3752
- Could not allow the JS fetch call for `merchant validation` go all the way to the Node endpoint
  - This involves certificates and difficult to stub the backend side of the call.
  - Therefore, had do use a `cy.intercept` when making the JS fetch call to `merchant validation`.

Google Pay Cypress tests
- Had to update the tests to use then connector stubs when authourising an Apple or Google payment.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -247,21 +247,21 @@
         "filename": "test/cypress/integration/web-payments/apple-pay.test.cy.js",
         "hashed_secret": "321f58f0cffa4caf8de5176503e82691f0f4f092",
         "is_verified": false,
-        "line_number": 33
+        "line_number": 34
       },
       {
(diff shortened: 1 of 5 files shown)


iqbalgds:
No longer required

PR #3771, opened 2023-11-17
PP-11039 System error nunjucks file - update page title

iqbalgds:
- Make the page title and H1 heading match - as this was a WCAG failure.
- This is the only error page that is affected by this.


SandorArpa (approved):
LOKTM

PR #3788, opened 2024-01-18
PP-11850 Update CSP > report-to

iqbalgds:
- Missed out the `Reporting-Endpoints` header.
- Required slightly rewriting how we specify the CSP in Helmet.


hjvoid (approved):
I'm confident that this change does not require further review and meets the secure coding requirements of [GOV.UK](http://gov.uk/) Pay

PR #3789, opened 2024-01-19
Add endpoint JSON info

hjvoid:
## WHAT
Update csp.js to include a JSON object containing csp endoint info as per the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-to).

### DIFF
--- a/app/middleware/csp.js
+++ b/app/middleware/csp.js
@@ -52,8 +52,16 @@ const connectSourceCardDetails = ['\'self\'', 'https://www.google-analytics.com'
 
 const skipSendingCspHeader = (req, res, next) => { next() }
 
-const setResponseCspReportingEndpointName = (req, res, next) => {
-  res.setHeader('Reporting-Endpoints', `${CSP_REPORTING_ENDPOINT_NAME}="${paths.csp.path}"`)
+const cspEndpointJson = {
+  group: CSP_REPORTING_ENDPOINT_NAME,
+  max_age: 100000,
+  endpoints: [
+    { url: `${paths.csp.path}` }
+  ]
+}
+
+const setResponseCspReportingEndpoint = (req, res, next) => {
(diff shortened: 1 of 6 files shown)


app/middleware/csp.js
```diff
@@ -52,8 +52,16 @@ const connectSourceCardDetails = ['\'self\'', 'https://www.google-analytics.com'
...
+const cspEndpointJson = {
+  group: CSP_REPORTING_ENDPOINT_NAME,
+  max_age: 100000,
+  endpoints: [
+    { url: `${paths.csp.path}` }
+  ]
+}
+
+const setResponseCspReportingEndpoint = (req, res, next) => {
+  res.setHeader('Report-To', cspEndpointJson)
```
iqbalgds, line 64: Might need to Stringify the JSON object

server.js
```diff
@@ -189,6 +189,7 @@ function initialise () {
   if (NODE_ENV !== 'test') {
     app.use(metrics.initialise())
   }
+  app.use(setReportingEndpoints) // adds reporting-endpoints header to all responses
```
iqbalgds, line 192: Nice

test/middleware/csp.test.js
```diff
@@ -177,7 +177,7 @@ describe('CSP middleware', () => {
     const csp = requireHelper('../../app/middleware/csp')
 
     const next = sinon.spy()
-    const response = { setHeader: sinon.spy() }
+    const response = { setHeader: sinon.spy(), removeHeader: sinon.spy() }
```
iqbalgds, line 180: Just wondering why we need to add a spy for this function?

nlsteers: the helmet upgrade, presume it now removes headers in some cases and this was required to make the tests green

PR #3791, opened 2024-01-22
PP-11850 Reverting CSP change that was causing Naxsi issues

iqbalgds:
- Removing the code in the content security policy that added code for `report-to`
- Was causing an Naxsi issue.


iqbalgds:
Closing - will raise another PR.

PR #3792, opened 2024-01-24
PP-12084 Do not show wallet options for moto payments

james-peacock-gds:
- If service is moto, do not show wallet payment options

### DIFF
--- a/app/views/charge.njk
+++ b/app/views/charge.njk
@@ -41,7 +41,7 @@
 
 
 {% block bodyStart %}
-  {% if (allowApplePay or allowGooglePay) and not savePaymentInstrumentToAgreement %}
+  {% if (allowApplePay or allowGooglePay) and not savePaymentInstrumentToAgreement and not moto%}
     <script nonce="{{ nonce }}">
       {% if allowApplePay %}
         if (window.ApplePaySession && window.ApplePaySession.canMakePayments() && window.ApplePaySession.supportsVersion(4)) {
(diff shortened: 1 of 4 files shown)


app/views/charge.njk
```diff
@@ -41,7 +41,7 @@
 
 
 {% block bodyStart %}
-  {% if (allowApplePay or allowGooglePay) and not savePaymentInstrumentToAgreement %}
+  {% if (allowApplePay or allowGooglePay) and not savePaymentInstrumentToAgreement and not moto%}
```
stephencdaly, line 44: Can you just fix the formatting to add a space before ` %}`

test/cypress/integration/web-payments/google-pay.test.cy.js
```diff
@@ -326,6 +329,15 @@ describe('Google Pay payment flow', () => {
       cy.log('Should not show Google Pay as payment is a recurring one')
 
       cy.get('#google-pay-payment-method-submit.web-payment-button--google-pay').should('not.exist')
+
+      cy.task('clearStubs')
+      cy.task('setupStubs', [...chargeStubsWithGooglePayOrApplePayEnabled(true, false, true, true)])
```
stephencdaly, line 334: Interesting that this is part of the same test rather than a separate one... You're just following the pattern that already existed though.

stephencdaly (commented):
Can you just fix the formatting in the nunjucks file, other than that looks good.

app/views/charge.njk
james-peacock-gds: thanks, done.

test/cypress/integration/web-payments/google-pay.test.cy.js
james-peacock-gds: yes, not sure why that was, but followed this format rather than refactor the whole thing

stephencdaly (approved):

PR #3794, opened 2024-02-02
NO-OP to test workflow fix

jfharden:
## WHAT

## HOW 
_Steps to test or reproduce:_


jfharden:
This is good, it has successfully validated the changes to the reusable manifest check workflow work as expected, and are less janky than the old way.

PR #3797, opened 2024-02-07
PP-11850 enclose reporting endpoint url in quotes

nlsteers:
## WHAT

🤦🏻‍♂️

### DIFF
--- a/app/middleware/csp.js
+++ b/app/middleware/csp.js
@@ -90,7 +90,7 @@ const worldpayIframeCSP = helmet({
 })
 
 const setReportingEndpoints = (req, res, next) => {
-  res.setHeader('Reporting-Endpoints', `${reportingEndpointName}=${frontendUrl}${paths.csp.path}`)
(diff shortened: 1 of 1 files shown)


alexbishop1 (approved):
The `Reporting-Endpoints` header is defined at:

https://w3c.github.io/reporting/#header

This says its value is an `sf-dictionary`.

The `sf-dictionary` data type is defined at:

https://www.rfc-editor.org/rfc/rfc8941.html#section-3.2

This says that the bit after the `=` is a `member-value`, which in this case is an `sf-item`.

The `sf-item` data type is defined here:

https://www.rfc-editor.org/rfc/rfc8941.html#section-3.3

An `sf-item` can be various things; the only one that makes sense for a URL is an `sf-string`.

The `sf-string` data type is defined here:

https://www.rfc-editor.org/rfc/rfc8941.html#section-3.3.3

This says that strings need to be surrounded with double quotation marks:

```
sf-string = DQUOTE *chr DQUOTE
```

PR #3800, opened 2024-02-12
BAU increase csp request size limit

nlsteers:
## WHAT

- Chrome Reporting API can batch csp violation reports prior to sending, leading to larger than expected payload sizes. To ensure we are receiving all reports we need to increase the max payload size on the `/csp-report` endpoint


nlsteers:
requires https://github.com/alphagov/pay-infra/pull/4733

jfharden (approved):

PR #3804, opened 2024-02-20
PP-12248 Log error message when rendering error page

stephencdaly:
If we catch an error and render the error page, we do not currently log the error message. The `error` always evaluates as an empty JSON object when we attempt to log it so it is not included. Instead, log the error message.


SandorArpa (approved):
LGTM

PR #3807, opened 2024-02-29
PP-12327 Bump govuk-frontend to 4.8.0

iqbalgds:
- For some reason, pay-frontend was not the latest version of govuk-frontend v4.
- It was using 4.7.8 instead of 4.8.0.
- This PR bumps govuk-frontend to 4.8.0.


SandorArpa (approved):
LGTM

PR #3822, opened 2024-04-12
PP-11683 Update healthcheck controller to use Axios

iqbalgds:
- Uses the Axios base client from `pay-js-commons`.


JFSGDS (approved):
Looks good.

PR #3823, opened 2024-04-15
PP-11682 Replace 'request' package with 'axios' for 'cardid.client'.

JFSGDS:
https://payments-platform.atlassian.net/browse/PP-11682

## WHAT
- Add axios-base-client from latest pay-js-commons.
- Import config files.
- Remove ‘base.client’ reference.

## HOW 

- All automated tests should pass preferably without the need to be amended.
- Existing functionality should remain unaffected.

### DIFF
--- a/app/models/card.js
+++ b/app/models/card.js
@@ -29,28 +29,28 @@ const checkCard = function (cardNo, allowed, blockPrepaidCards, language, correl
     }
     cardIdClient.post({ payload: data, correlationId: correlationId })
       .then((response) => {
-        incrementStatusCodeCounter('checkCard', response.statusCode)
+        incrementStatusCodeCounter('checkCard', response.status)
         logger.info('POST to %s ended - total time %dms', cardIdClient.CARD_URL, new Date() - startTime, loggingFields)
 
-        if (response.statusCode === 404) {
+        if (response.status === 404) {
           return reject(new Error('Your card is not supported'))
         }
-        if (response.statusCode === 422) {
+        if (response.status === 422) {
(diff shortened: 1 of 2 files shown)


app/models/card.js
```diff
@@ -29,28 +29,29 @@ const checkCard = function (cardNo, allowed, blockPrepaidCards, language, correl
...
-            status_code: response.statusCode,
+            status_code: response.status,
             url: cardIdClient.CARD_URL
           })
           return resolve()
         }
 
-        const body = response.body
+        const body = response.data
+        // const body = response.body
```
iqbalgds, line 54: Just need to remove this

iqbalgds (changes requested):
1 very minor comment

app/services/clients/cardid.client.js
```diff
@@ -1,16 +1,27 @@
...
 /**
  *
  * @param args
  *
  * @returns {Request}
  */
-exports.post = (args) => baseClient.post(CARD_URL, args, null)
+
+exports.post = async (args) => {
+  try {
```
iqbalgds, line 20: We don't need a try / catch here.

The calling method wraps this in a promise anyway and will deal with the error.

iqbalgds (changes requested):
Just a minor thing.

app/models/card.js
JFSGDS: Good spot. Done.

app/services/clients/cardid.client.js
iqbalgds: As discussed, this is no longer required.
As Axios behaves a bit differently to request library.

PR #3841, opened 2024-05-30
Changes for pay-frontend to use axios via pay-js-commons

marcotranchino:
With this PR, we are proposing two changes to pay-frontend in order to use
the new axios client via pay-js-commons.

The changes are only in the final two commits listed below, just after the 
[temporary commit ](https://github.com/alphagov/pay-frontend/pull/3841/commits/bd3a06b8b4e253ac77a451f94284d5a87bfa0b6b) to use the local (WIP) version of pay-js-commons. 

[The first commit](https://github.com/alphagov/pay-frontend/commit/51c9b20f3bb19e82365d513b8536cce1c8ce0184) checks explicitly for success status codes returned by the 
client and throws CLIENT_UNAVAILABLE for all other codes.

[The second commit ](https://github.com/alphagov/pay-frontend/commit/665a7b559b160fee295f2b3363678c2db031ab06)ensures that the test is setup correctly
by replying with a status code that is not included in the options handled by the client, so that we can verify
that the operation is managed correctly by resolving the promise.

This is all to be discussed with @iqbalgds.

### DIFF
--- a/app/models/token-axios.js
+++ b/app/models/token-axios.js
@@ -1,0 +1,22 @@
+'use strict'
+
+// Local dependencies
+const connectorClient = require('../services/clients/connector-axios.client')
+
+const markTokenAsUsed = async function (tokenId, loggingFields = {}) {
+  let response
+  try {
+    response = await connectorClient().markTokenAsUsed({ tokenId }, loggingFields)
+  } catch (err) {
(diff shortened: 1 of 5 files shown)


iqbalgds:
Closing - as this has been superceeded by newer commits.

PR #3843, opened 2024-06-05
PP-11682 Fix admin client PACT test

iqbalgds:
- These tests were not running properly, they were using a timeout and inconsistent with the other PACT test.
- Therefore, removed the timeout and made it consistent.


hjvoid (approved):
LGTM 😎

PR #3849, opened 2024-06-25
PP-11682 - Update axios connect tests

iqbalgds:
- Update tests to work with the updated `axios-base-client` from pay-js-commons.
- All status codes are now considered successful.  The consuming code will now need to decide how to manage not 2xx responses.


SandorArpa (approved):
LGTM

PR #3864, opened 2024-07-15
[deploying to test only] PP-12853 Update apple pay merchant validation to use Axios

marcotranchino:
With this change, we are removing the use of `requestretry` from the Merchant Validation Controller which currently uses this module.

We are replacing it by using Axios directly.

Note that the previous attempted did not pass our QA on the test environment[1].

Tests on the local environment are successful, so we are adding logs for the test environment.

Further information in Jira[2].

[1]
https://github.com/alphagov/pay-frontend/pull/3862

[2]


iqbalgds (approved):

SandorArpa (approved):
LOKTM

PR #3866, opened 2024-07-16
PP-12161 Prevent duplication of IDs within payment summary

marcotranchino:
With this change, we are updating the payment summary macro so that when it is
called for a second time it sets its elements with different IDs.

This allows us also to enable a test to check when Apple Pay is not supported
by the browser, which was previously created in a separate branch and
commented out as it would not work given the repetition of IDs caused by the
payment summary macro.

Further information in Jira.

https://payments-platform.atlassian.net/browse/PP-12161

### DIFF
--- a/test/cypress/integration/web-payments/apple-pay.test.cy.js
+++ b/test/cypress/integration/web-payments/apple-pay.test.cy.js
@@ -258,6 +258,10 @@ describe('Apple Pay payment flow', () => {
 
       // 8. User should see normal payment form
       cy.get('#card-no').should('be.visible')
+
+      // 9. User should see the payment summary for the wallet option
+      cy.get('#amount').should('be.hidden')
+      cy.get('#wallet-amount').should('be.visible')
     })
 
     it('Should show Apple Pay but error because a bad email was entered', () => {
@@ -324,8 +328,11 @@ describe('Apple Pay payment flow', () => {
 
       // 8. User should see normal payment form
       cy.get('#card-no').should('be.visible')
-    })
 
+      // 9. User should see the payment summary for the non-wallet option
+      cy.get('#amount').should('be.visible')
+      cy.get('#wallet-amount').should('not.exist')
+    })
 
     it('Should not show Apple Pay when enabled but browser does not support it', () => {
       cy.task('setupStubs', createPaymentChargeStubs)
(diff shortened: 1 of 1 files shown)


app/views/charge.njk
```diff
@@ -18,22 +18,22 @@
...
+{% macro paymentSummary(main = true) %}
   <div class="payment-summary-wrap">
     <div class="payment-summary">
         <h2 class="govuk-heading-m">{{ __p("paymentSummary.title") }}</h2>
-        <p id="payment-description" class="govuk-body">
+        <p id="{{ 'payment-description' if main else 'secondary-payment-description' }}" class="govuk-body">
         {{ description }}
         </p>
-        <p class="govuk-body hidden" id="payment-summary-breakdown">
+        <p class="govuk-body hidden" id="{{ 'payment-summary-breakdown' if main else 'secondary-payment-summary-breakdown' }}">
```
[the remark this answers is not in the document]
iqbalgds: I think its fine to leave as it is for now

app/views/charge.njk
```diff
@@ -18,22 +18,22 @@
 
 {% set shouldShowEmail = (gatewayAccount.emailCollectionMode === 'MANDATORY') or (gatewayAccount.emailCollectionMode === 'OPTIONAL') %}
 
-{% macro paymentSummary() %}
+{% macro paymentSummary(main = true) %}
   <div class="payment-summary-wrap">
     <div class="payment-summary">
         <h2 class="govuk-heading-m">{{ __p("paymentSummary.title") }}</h2>
-        <p id="payment-description" class="govuk-body">
+        <p id="{{ 'payment-description' if main else 'secondary-payment-description' }}" class="govuk-body">
```
iqbalgds, line 25: Instead of using the word secondary - can we use `wallet` instead?
E.g. `wallet-payment-description`
Please let me know what you think?

test/cypress/integration/web-payments/apple-pay.test.cy.js
```diff
@@ -326,6 +326,33 @@ describe('Apple Pay payment flow', () => {
       cy.get('#card-no').should('be.visible')
     })
 
+
+    it('Should not show Apple Pay when enabled but browser does not support it', () => {
```
iqbalgds, line 330: Not sure if my thinking is correct - can we add new tests to check for the new CSS classes?
Would be good to check each one to make sure it's def correct.

iqbalgds (changes requested):
Looks really good.
Just got a few ideas - please let me know what you think

iqbalgds (approved):
Good work!

PR #3867, opened 2024-07-17
PP-11682 - Remove old base client files

iqbalgds:
- These have been superceeded by the `axios-base-client` that is imported from `pay-js-commons`


alexbishop1 (approved):
🚮

PR #3868, opened 2024-07-18
PP-12546 hold payments in state after completion

nlsteers:
## WHAT

frontend now holds up to 10 charge objects on the client to allow users
who navigate back to Pay from the service to see the status of their payment

a charge will be removed from the client if 10 charges exist and an 11th payment
is started

the logic for determining which charge to remove is documented in this ADR:
https://github.com/alphagov/pay-architecture/blob/main/adr/026-keep-payments-in-cookie-for-longer.md

- new ChargeState model
- new csrf secret management middleware
  - csrf secret is set at the root of the session object as part of a gradual migration of csrf behaviour
- remove used csrf token store from frontend state
- add new cookie functions for handling charge state
- remove unused session code from server

### DIFF
--- a/app/controllers/return.controller.js
+++ b/app/controllers/return.controller.js
@@ -6,15 +6,24 @@ const Charge = require('../models/charge')
 const responseRouter = require('../utils/response-router')
 const { CORRELATION_HEADER } = require('../../config/correlation-header')
 const { withAnalyticsError } = require('../utils/analytics')
-const cookies = require('../utils/cookies')
 const { createChargeIdSessionKey } = require('../utils/session')
+const cookies = require('../utils/cookies')
 
 exports.return = (req, res) => {
   const correlationId = req.headers[CORRELATION_HEADER] || ''
   const charge = Charge(correlationId)
-  // Remove the charge data from the cookie
-  const cookieKey = createChargeIdSessionKey(req.chargeId)
-  cookies.deleteSessionVariable(req, cookieKey)
+  const chargeKey = createChargeIdSessionKey(req.chargeId)
+  const chargeState = cookies.getSessionChargeState(req, chargeKey)
+  if (chargeState) {
+    chargeState.markTerminal()
+    cookies.setSessionChargeState(req, chargeKey, chargeState)
+  } else {
+    /* TODO: remove after PP-12546 has been merged
+    /  we no longer need to delete the session variable in the return controller as the cookie is cleaned up in the
+    /  secure controller, we're just cleaning up any existing payment journeys prior to the merge of PP-12546
+    */
+    cookies.deleteSessionVariable(req, chargeKey)
+  }
   if (charge.isCancellableCharge(req.chargeData.status)) {
     return charge.cancel(req.chargeId, getLoggingFields(req))
       .then(() => logger.warn('Return controller cancelled payment', getLoggingFields(req)))
--- a/app/controllers/secure.controller.js
+++ b/app/controllers/secure.controller.js
@@ -1,7 +1,7 @@
 'use strict'
 
-// NPM dependencies
 const csrf = require('csrf')
+
 const {
   PAYMENT_EXTERNAL_ID,
   GATEWAY_ACCOUNT_ID,
@@ -16,11 +16,12 @@ const Token = require('../models/token')
 const Charge = require('../models/charge')
 const responseRouter = require('../utils/response-router')
 const { createChargeIdSessionKey } = require('../utils/session')
-const { setSessionVariable, getSessionVariable } = require('../utils/cookies')
+const cookies = require('../utils/cookies')
 const CORRELATION_HEADER = require('../../config/correlation-header').CORRELATION_HEADER
 const withAnalyticsError = require('../utils/analytics').withAnalyticsError
 const { resolveActionName } = require('../services/state.service')
 const paths = require('../paths')
+const { ChargeState } = require('../models/ChargeState')
 
 exports.new = async function (req, res) {
   const chargeTokenId = req.params.chargeTokenId || req.body.chargeTokenId
@@ -36,7 +37,7 @@ exports.new = async function (req, res) {
     setLoggingField(req, GATEWAY_ACCOUNT_TYPE, tokenResponse.charge.gateway_account.type)
 
     if (tokenResponse.used === true) {
-      if (!getSessionVariable(req, createChargeIdSessionKey(chargeId))) {
+      if (!cookies.getSessionVariable(req, createChargeIdSessionKey(chargeId))) {
         throw new Error('UNAUTHORISED')
       }
       logger.info('Payment token being reused', getLoggingFields(req))
@@ -48,7 +49,12 @@ exports.new = async function (req, res) {
     } else {
       logger.info('Payment token used for the first time', getLoggingFields(req))
       await Token.markTokenAsUsed(chargeTokenId, correlationId, getLoggingFields(req))
-      setSessionVariable(req, createChargeIdSessionKey(chargeId), { csrfSecret: csrf().secretSync() })
+      if (cookies.getChargesOnSession(req).length >= 10) {
+        const chargeToDelete = cookies.findSessionChargeToDelete(req)
+        cookies.deleteSessionVariable(req, chargeToDelete)
+      }
+      cookies.setSessionChargeState(req, createChargeIdSessionKey(chargeId), new ChargeState())
+      cookies.setSessionVariable(req, createChargeIdSessionKey(chargeId), { csrfSecret: csrf().secretSync() }) // TODO: remove after PP-12546 has been merged
       res.redirect(303, generateRoute(resolveActionName(chargeStatus, 'get'), { chargeId }))
     }
   } catch (err) {
--- a/app/middleware/csrf.js
+++ b/app/middleware/csrf.js
@@ -8,11 +8,21 @@ const logger = require('../utils/logger')(__filename)
 const { getLoggingFields } = require('../utils/logging-fields-helper')
 const session = require('../utils/session')
 const responseRouter = require('../utils/response-router')
+const cookies = require('../utils/cookies')
+
+exports.csrfSetSecret = (req, _, next) => {
+  const csrfSecret = cookies.getSessionCsrfSecret(req)
+  if (!csrfSecret) {
+    logger.info('Setting CSRF secret for session')
+    cookies.setSessionVariable(req, 'csrfSecret', csrf().secretSync())
+  }
+  next()
+}
 
 exports.csrfTokenGeneration = (req, res, next) => {
   const chargeId = fetchAndValidateChargeId(req)
   const chargeSession = session.retrieve(req, chargeId)
-  res.locals.csrf = csrf().create(chargeSession.csrfSecret)
+  res.locals.csrf = csrf().create(chargeSession.csrfSecret) // TODO: generate tokens using the session secret
   next()
 }
 
@@ -23,11 +33,11 @@ exports.csrfCheck = (req, res, next) => {
     return responseRouter.response(req, res, 'UNAUTHORISED')
   }
 
-  const chargeSession = session.retrieve(req, chargeId) || {}
+  const chargeSession = session.retrieve(req, chargeId) || {} // TODO: remove after PP-12546 has been merged
+  const sessionCsrfSecret = cookies.getSessionCsrfSecret(req)
   const csrfToken = req.body.csrfToken
-  chargeSession.csrfTokens = chargeSession.csrfTokens || []
 
-  if (!chargeSession.csrfSecret) {
+  if (!chargeSession.csrfSecret && !sessionCsrfSecret) {
     responseRouter.response(req, res, 'UNAUTHORISED')
     logger.warn('CSRF secret is not defined', {
       ...getLoggingFields(req),
@@ -35,10 +45,9 @@ exports.csrfCheck = (req, res, next) => {
       url: req.originalUrl,
       method: req.method
     })
-  } else if (!csrfValid(csrfToken, chargeSession, req)) {
+  } else if (!csrfValid(csrfToken, chargeSession.csrfSecret, req) && !csrfValid(csrfToken, sessionCsrfSecret, req)) {
     responseRouter.systemErrorResponse(req, res, 'CSRF is invalid')
   } else {
-    chargeSession.csrfTokens.push(csrfToken)
     next()
   }
 }
@@ -50,13 +59,13 @@ function fetchAndValidateChargeId (req) {
   return false
 }
 
-function csrfValid (csrfToken, chargeSession, req) {
+function csrfValid (csrfToken, secret, req) {
+  if (!secret) {
+    return false
+  }
   if (!['put', 'post'].includes(req.method.toLowerCase())) {
     return true
-  } else if (chargeSession.csrfTokens.includes(csrfToken)) {
-    logger.warn('CSRF token was already used', getLoggingFields(req))
-    return false
   } else {
-    return csrf().verify(chargeSession.csrfSecret, csrfToken)
+    return csrf().verify(secret, csrfToken)
   }
 }
--- a/app/models/ChargeState.js
+++ b/app/models/ChargeState.js
@@ -1,0 +1,52 @@
+const logger = require('../utils/logger')(__filename)
+
+class ChargeState {
+  /**
+   * @param {Number} [createdAt]
+   * @param {Number} [accessedAt]
+   * @param {boolean} [isTerminal]
+   */
+  constructor (createdAt, accessedAt, isTerminal) {
+    this.createdAt = createdAt || epochSecondsNow()
+    this.accessedAt = accessedAt || epochSecondsNow()
+    this.isTerminal = isTerminal || false
+  }
+
+  updateAccessedAt () {
+    this.accessedAt = epochSecondsNow()
+  }
+
+  markTerminal () {
+    this.updateAccessedAt()
+    this.isTerminal = true
+  }
+
+  toString () {
+    return `${this.createdAt},${this.accessedAt},${this.isTerminal ? 'T' : 'F'}`
+  }
+}
+
+const chargeStateFromString = (data) => {
+  try {
+    const dataParts = data.split(',')
+    if (dataParts.length === 3) {
+      return new ChargeState(Number(dataParts[0]), Number(dataParts[1]), dataParts[2] === 'T')
+    } else {
+      logger.error('argument is not a valid ChargeState')
+      return null
+    }
+  } catch (e) {
+    logger.warn(`Error de-serialising ChargeState from string: ${e.message}`)
+    return null
+  }
+}
+
+const epochSecondsNow = () => {
+  return Math.floor(Date.now() / 1000)
+}
+
+module.exports = {
+  ChargeState,
+  chargeStateFromString,
+  epochSecondsNow
+}
--- a/app/utils/cookies.js
+++ b/app/utils/cookies.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const clientSessions = require('client-sessions')
+const { chargeStateFromString, epochSecondsNow } = require('../models/ChargeState')
 const logger = require('../utils/logger')(__filename)
 
 // Constants
@@ -17,7 +18,11 @@ module.exports = {
   namedCookie,
   deleteSessionVariable,
   isSessionPresent,
-  getSessionVariableNames
+  getChargesOnSession,
+  getSessionCsrfSecret,
+  getSessionChargeState,
+  setSessionChargeState,
+  findSessionChargeToDelete
 }
 
 /**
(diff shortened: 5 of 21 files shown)


alexbishop1 (changes requested):
Thanks for taking the time to talk me through this earlier. This seems to implement what ADR 026 specifies, maintaining support for. existing cookies (though I do have a concern about CSRF for in-flight payments described below).

As discussed, we believe switching from per-charge CSRF secrets to per-user CSRF secrets does not make our CSRF protection less secure.

You’ve shown me https://github.com/alphagov/pay-endtoend/pull/1331, which proves that this change enables the behaviour we want of showing a friendly error message to users who try to return to a completed payment.

Most of the comments below are stylistic.

test/middleware/csrf.test.js
```diff
@@ -97,18 +99,11 @@ describe('retrieve param test', function () {
...
     csrfTokenGeneration(validGetRequest, resp, next)
     assertValidRequest(next, resp, status, render)
-    assert.strictEqual(validPost.frontend_state.ch_foo.csrfTokens[0], validPost.body.csrfToken)
-  })
-
-  it('should be unsuccessful on post if token is already used', function () {
-    const resp = _.cloneDeep(response)
-    csrfCheck(validPost, resp, next)
-    assertErrorRequest(next, resp, status, render)
   })
```
nataliecarey, line 107: I'm nervous about the removal of the csrf tests when that logic's changed (written before getting to the end - maybe they've been moved into a different file)

test/unit/cookies.test.js
```diff
@@ -122,6 +123,138 @@ describe('cookie configuration', function () {
...
 
+describe('reduce session charge count', () => {
+  let initialEnvironmentVariables
+  before(() => {
+    initialEnvironmentVariables = Object.assign({}, process.env)
+  })
+
+  afterEach(() => {
+    for (const envVar in process.env) {
+      process.env[envVar] = initialEnvironmentVariables[envVar]
```
nataliecarey, line 134: This doesn't quite reset them to their previous state, I'd suggest checking `initialEnvironmentVariables.hasOwnProperty(envVar)` and if that's false running a `delete proces.env[envVar]`

nataliecarey (commented):
I've added some comments, I also agree with @alexbishop1's comments.

test/controllers/return.controller.test.js
```diff
@@ -57,13 +67,15 @@ describe('return controller', function () {
...
-      .post('/v1/frontend/charges/aChargeId/cancel')
+      .post(`/v1/frontend/charges/${chargeId}/cancel`)
       .reply(204)
 
     requireReturnController().return(request, response)
       .should.be.fulfilled.then(() => {
         assert(response.redirect.calledWith('http://a_return_url.com'))
       }).should.notify(done)
+    const chargeStateUnderTest = chargeStateFromString(request.frontend_state[`ch_${chargeId}`].data)
+    expect(chargeStateUnderTest.isComplete).to.equal(true)
```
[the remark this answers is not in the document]
nlsteers: I think you might have misread the code, apologies if it's not clear. The `beforeEach` sets the charge session data to `new ChargeState().toString()`. The class constructor defaults `isComplete` to `false` if no arguments are provided.

This test is asserting that a change has been made (charge state `isComplete` has been set to `true`, the expected behaviour of the return controller).

app/models/ChargeState.js
```diff
@@ -0,0 +1,40 @@
+class ChargeState {
+  /**
+   * @param {Number} [createdAt]
+   * @param {Number} [accessedAt]
+   * @param {string} [isComplete] (T or F)
```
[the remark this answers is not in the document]
nlsteers: Yes I think I agree, will change

test/middleware/csrf.test.js
nlsteers: this test has been removed because we no longer store used csrf tokens.

test/unit/cookies.test.js
nlsteers: interesting, this was copied from another test so we'll need to revisit other places this is used

test/controllers/secure.controller.test.js
```diff
@@ -84,9 +82,51 @@ describe('secure controller', function () {
...
+              ch_10: {
+                data: chargeState.toString()
+              },
+            },
+            params: { chargeTokenId: 1 },
+            headers: { 'x-Request-id': 'unique-id' }
+          }
+          await requireSecureController(getFindByTokenSuccessStub(false, 'CREATED'), true, responseRouter).new(requestWithTenChargesOnCookie, response)
+          expect(response.redirect.calledWith(303, paths.generateRoute('card.new', {chargeId: chargeId}))).to.be.true // eslint-disable-line
+          expect(requestWithTenChargesOnCookie.frontend_state).to.have.all.keys('ch_' + chargeId, 'ch_1','ch_2','ch_3','ch_4','ch_5','ch_6','ch_7','ch_8','ch_10',)
```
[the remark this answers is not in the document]
nlsteers: yes I'll make it clearer

app/utils/session.js
```diff
@@ -21,28 +22,44 @@ const validateSessionCookie = function validateSessionCookie (req) {
...
+    const chargeState = chargeStateFromString(stateString)
+    chargeState.updateAccessedAt()
+    cookies.setSessionChargeState(req, createChargeIdSessionKey(chargeId), chargeState)
+  } else {
+    /* TODO remove this else condition after a few days (lol will this ever happen)
+    / this is here to support any existing payment journeys prior to the merge of PP-12546
+    */
+    logger.warn('ChargeId found on session but charge state was not found', {
+      ...getLoggingFields(req)
+    })
```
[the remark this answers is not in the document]
nlsteers: warn seemed appropriate given it's not expected behaviour for any future payments and should be removed at some point.

app/controllers/secure.controller.js
```diff
@@ -48,7 +47,12 @@ exports.new = async function (req, res) {
     } else {
       logger.info('Payment token used for the first time', getLoggingFields(req))
       await Token.markTokenAsUsed(chargeTokenId, correlationId, getLoggingFields(req))
-      setSessionVariable(req, createChargeIdSessionKey(chargeId), { csrfSecret: csrf().secretSync() })
+      if (cookies.getChargesOnSession(req).length >= 10) {
+        logger.info(`[${cookies.getSessionCookieName()}] contains 10 charges`)
```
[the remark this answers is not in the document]
nlsteers: the session utility already logs the number of charges on the session for every request on the standard middleware stack, I can add it here as well if you prefer

app/controllers/secure.controller.js
```diff
@@ -48,7 +47,12 @@ exports.new = async function (req, res) {
...
+        const chargeToRemove = cookies.reduceSessionChargeCount(req)
+        cookies.deleteSessionVariable(req, chargeToRemove)
+      }
+      cookies.setSessionChargeState(req, createChargeIdSessionKey(chargeId), new ChargeState())
```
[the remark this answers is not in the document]
nlsteers: I felt that the `new` keyword conveyed that intent but happy to amend if you feel strongly about it

app/controllers/return.controller.js
```diff
@@ -6,15 +6,26 @@ const Charge = require('../models/charge')
...
+const cookies = require('../utils/cookies')
 
 exports.return = (req, res) => {
   const correlationId = req.headers[CORRELATION_HEADER] || ''
   const charge = Charge(correlationId)
-  // Remove the charge data from the cookie
-  const cookieKey = createChargeIdSessionKey(req.chargeId)
-  cookies.deleteSessionVariable(req, cookieKey)
+  const chargeKey = createChargeIdSessionKey(req.chargeId)
+  const stateString = cookies.getSessionChargeState(req, chargeKey)
```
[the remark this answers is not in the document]
nlsteers: ✅

app/middleware/csrf.js
```diff
@@ -23,22 +32,20 @@ exports.csrfCheck = (req, res, next) => {
     return responseRouter.response(req, res, 'UNAUTHORISED')
   }
 
-  const chargeSession = session.retrieve(req, chargeId) || {}
+  const csrfSecret = cookies.getSessionCsrfSecret(req)
```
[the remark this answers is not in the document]
nlsteers: I've added some backwards compatibility logic to look for the CSRF secret in the old location (charge object) if the lookup on session object fails

app/models/ChargeState.js
```diff
@@ -0,0 +1,40 @@
...
+    this.createdAt = createdAt || epochSecondsNow()
+    this.accessedAt = accessedAt || epochSecondsNow()
+    this.isComplete = isComplete === 'T'
+  }
+
+  updateAccessedAt () {
+    this.accessedAt = epochSecondsNow()
+  }
+
+  done () {
```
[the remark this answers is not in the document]
nlsteers: I've updated to `markTerminal` as complete felt too 'successful' given that completed payments can be cancelled/declined etc.

app/utils/cookies.js
```diff
@@ -133,6 +138,54 @@ function deleteSessionVariable (req, key) {
...
+
+  if (completedCharge) return completedCharge.chargeId
+
+  const olderThanNinetyMinutesCharge = chargesSortedByAccessedDate.filter(charge => {
+    return charge.state.createdAt < ninetyMinutesAgo
+  })[0]
+
+  if (olderThanNinetyMinutesCharge) return olderThanNinetyMinutesCharge.chargeId
+
+  return chargesSortedByAccessedDate[0].chargeId
```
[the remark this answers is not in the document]
nlsteers: agreed, have added some logging to surface this

app/controllers/secure.controller.js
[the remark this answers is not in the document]
nlsteers: I've added logging to `findSessionChargeToDelete` that logs the number on the session object

test/models/chargestate.test.js
```diff
@@ -0,0 +1,59 @@
+const expect = require('chai').expect
+const sinon = require('sinon')
+const { ChargeState, chargeStateFromString } = require('../../app/models/ChargeState')
+
+describe('ChargeState model should', () => {
+  let clock
+
+  beforeEach(() => {
+    let fixedDate = new Date(1721398795348) // 19/07/2024 15:20
```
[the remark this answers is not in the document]
nlsteers: have updated with easier to grok numbers

app/utils/cookies.js
```diff
@@ -163,12 +228,22 @@ function getSessionVariable (req, key) {
   return session && session[key]
 }
 
+function getSessionChargeState (req, key) {
+  return getSessionVariable(req, key).data
```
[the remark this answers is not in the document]
nlsteers: the call sites for getSessionChargeState have handling code for if the returned value is `undefined`, this may be the case for any existing charges in flight prior to the merge of this PR

app/utils/cookies.js
```diff
@@ -133,6 +138,54 @@ function deleteSessionVariable (req, key) {
...
+    })
+    .sort(({ state: a }, { state: b }) => a.accessedAt - b.accessedAt)
+
+  const olderThanTwentyFourHoursCharge = chargesSortedByAccessedDate.filter(charge => {
+    return charge.state.createdAt < yesterday
+  })[0]
+
+  if (olderThanTwentyFourHoursCharge) return olderThanTwentyFourHoursCharge.chargeId
+
+  const completedCharge = chargesSortedByAccessedDate.filter(charge => {
```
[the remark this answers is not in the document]
nlsteers: 👍🏻 , have replaced all `.filter` with `.find`

test/unit/cookies.test.js
```diff
@@ -122,6 +123,138 @@ describe('cookie configuration', function () {
...
+          data: new ChargeState(twentyMinutesAgo, (now - 45), 'F').toString()
+        },
+        ch_3: {
+          data: new ChargeState(twentyMinutesAgo, (now - 30), 'F').toString()
+        },
+        ch_4: {
+          data: new ChargeState(twentyMinutesAgo, (now - 40), 'F').toString()
+        }
+      }
+    }
```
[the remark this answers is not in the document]
nlsteers: I specifically wanted to give examples of differently shaped session objects to provide more resolution on the behaviour of the `findSessionChargeToDelete` function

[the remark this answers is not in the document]
alexbishop1: Fair enough.

app/utils/cookies.js
```diff
@@ -133,6 +138,54 @@ function deleteSessionVariable (req, key) {
...
+ *
+ * created over 1 day ago
+ * has the completed flag set to true
+ * created over 90 minutes ago
+ * has the earliest last accessed date
+ *
+ * @param {Request} req
+ * @returns {string}
+ */
+function reduceSessionChargeCount (req) {
```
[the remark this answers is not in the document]
alexbishop1: `findSessionChargeToDelete` is also fine.

app/middleware/csrf.js
```diff
@@ -23,22 +32,21 @@ exports.csrfCheck = (req, res, next) => {
...
+    ?? session.retrieve(req, chargeId)['csrfSecret'] // TODO back compat for in-flight payments pre PP-12546, remove nullish coalescing op after merge
```
alexbishop1, line 36: I”m not sure this is going to be enough.

It solves the case of a user POSTing with a cookie that only has the CSRF secret inside the charge and not at the top level.

But consider the following case:

1. User hits a server that has been deployed with the new code and gets a new CSRF secret
2. User makes a POST that hits a server that’s still running the old code
3. Server running the old code gets the CSRF secret from inside the charge but it doesn’t match the newly-generated CSRF secret in step 1

I think we might have to make it so that the new top-level CSRF secret is stored within each charge too…

app/utils/cookies.js
```diff
@@ -133,6 +138,43 @@ function deleteSessionVariable (req, key) {
...
+        state: getSessionChargeState(req, chargeId)
+      }
+    })
+    .sort(({ state: a }, { state: b }) => a.accessedAt - b.accessedAt)
+
+  logger.info(`User session contains ${chargesSortedByAccessedDate.length} charges`)
+
+  const leastRecentlyAccessedChargeCreatedMoreThanTwentyFourHoursAgo = chargesSortedByAccessedDate.find(charge => charge.state.createdAt < yesterday)
+
+  if (leastRecentlyAccessedChargeCreatedMoreThanTwentyFourHoursAgo) { return leastRecentlyAccessedChargeCreatedMoreThanTwentyFourHoursAgo.chargeId }
```
alexbishop1, line 163: Egyption brackets, please:

```javascript
if (leastRecentlyAccessedChargeCreatedMoreThanTwentyFourHoursAgo) {
    leastRecentlyAccessedChargeCreatedMoreThanTwentyFourHoursAgo.chargeId
}
```

(Same goes for the lines below.)

app/utils/cookies.js
```diff
@@ -133,6 +138,43 @@ function deleteSessionVariable (req, key) {
...
+
+  const leastRecentlyAccessedTerminalCharge = chargesSortedByAccessedDate.find(charge => charge.state.isTerminal)
+
+  if (leastRecentlyAccessedTerminalCharge) { return leastRecentlyAccessedTerminalCharge.chargeId }
+
+  const leastRecentlyAccessedChargeCreatedMoreThanNinetyMinutesAgo = chargesSortedByAccessedDate.find(charge => charge.state.createdAt < ninetyMinutesAgo)
+
+  if (leastRecentlyAccessedChargeCreatedMoreThanNinetyMinutesAgo) { return leastRecentlyAccessedChargeCreatedMoreThanNinetyMinutesAgo.chargeId }
+
+  logger.info('No preferred charges found for deletion on session, defaulting to least recently accessed charge')
```
alexbishop1, line 173: Would logging the accessed timestamp be useful here?

alexbishop1 (changes requested):
Thanks for addressing the comments.

I’m still worried about CSRF and in-flight payments during the deployment itself (details in the comment below).

Other than, just one or two small things.

app/middleware/csrf.js
nlsteers: I think in this case frontend will throw an unauthorised error. The user would already have started a payment on a server not running the new code. Refresh the card entry page and hit a server running the new code (and generating a new secret and token), then POST to an old server with a token that will not be valid with the top level secret. 

There’s not really anything we can do about that if the user POSTs to an old server. 

We’re resolving for what will be an incredibly small number of payments (if any depending on when the PR is merged). 

Perhaps we should accept the chance of a very small number of payments being impacted and that this is a tolerable outcome. 

Or merge this change out of hours.

alexbishop1: > The user would already have started a payment on a server not running the new code.

Yes, that could happen. The user starts a payment on a server running the old code (so they get a per-charge CSRF secret) then hits a server running the new code (so they get a new top-level secret but the per-charge one is left alone), then hits a server running the old code again (which only knows how to read the old per-charge secrets).

However, I think it could happen in both cases If the user started a payment on a server running the new code (so they get the top-level CSRF secret but not the per-charge one) then hits a server running the old code (which only knows how to read the per-charge secrets). Since that involves fewer hops between old and new, it seems more likely.

Put together, I think we might end up breaking a significant number of the in-flight payments during deployment.

How hard would it be to make the code that writes the new top-level CSRF secret also write the same secret to each charge? I think that would solve it.

app/utils/cookies.js
nlsteers: done

.gitignore
```diff
@@ -36,3 +36,6 @@ pacts
 /test/cypress/screenshots
 /test/cypress/videos
 mb.pid
+
+# local apple pay
+app/assets/apple-pay/local
```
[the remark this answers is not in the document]
nlsteers: removed

app/middleware/csrf.js
nlsteers: resolving this comment, we've discussed the approach and are adopting a phased roll out of changes to blunt some of these sharp edges

app/utils/cookies.js
nlsteers: I don't think it's necessary, we can add more detail to the logs if we see this cropping up a lot. As it stands this should be quite rare

app/models/ChargeState.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  toString () {
+    return `${this.createdAt},${this.accessedAt},${this.isTerminal ? 'T' : 'F'}`
+  }
+}
+
+const chargeStateFromString = (data) => {
+  try {
+    const dataParts = data.split(',')
```
alexbishop1, line 31: What is this going to do if there’s no `data` (because it’s an old-format charge)?

app/models/ChargeState.js
```diff
@@ -0,0 +1,52 @@
...
+    if (dataParts.length === 3) {
+      return new ChargeState(Number(dataParts[0]), Number(dataParts[1]), dataParts[2] === 'T')
+    } else {
+      logger.error('argument is not a valid ChargeState')
```
alexbishop1, line 35: This is going to happen a lot at first but I suppose it’s OK as long as we’re aware.

app/utils/session.js
```diff
@@ -21,28 +21,43 @@ const validateSessionCookie = function validateSessionCookie (req) {
...
     return false
-  } else if (!cookie.getSessionVariable(req, 'ch_' + chargeId)) {
+  } else if (!cookies.getSessionVariable(req, chargeKey)) {
     logger.info('ChargeId was not found on the session', {
       ...getLoggingFields(req),
       referrer: req.get('Referrer'),
       url: req.originalUrl,
       method: req.method,
-      session_keys: cookie.getSessionVariableNames(req)
+      session_keys: cookies.getChargesOnSession(req)
```
alexbishop1, line 40: If I understand this right, this is only going to include charges that have a `data` property (i.e. use the new format). Is that going to be confusing?

app/utils/session.js
```diff
@@ -21,28 +21,43 @@ const validateSessionCookie = function validateSessionCookie (req) {
...
     })
     return false
   }
   logger.info('ChargeId found on session', {
     ...getLoggingFields(req),
-    number_of_payments_on_cookie: cookie.getSessionVariableNames(req).length
+    number_of_payments_on_cookie: cookies.getChargesOnSession(req).length
```
alexbishop1, line 46: Similarly, looks like this will only count new-format charges.

alexbishop1 (changes requested):
Thanks for revisiting this and I agree that scaling back the ambition of the initial CSRF changes is a good plan. I have some small concerns about some of the logging but otherwise I think it looks good.

app/models/ChargeState.js
nlsteers: There’s a test that covers this scenario. The function returns a null value that is handled at the call site.

alexbishop1: Right you are.

app/utils/session.js
alexbishop1: Sure thing.

alexbishop1 (approved):

PR #3877, opened 2024-07-26
PP-12853 Use axios or request retry for Apple Pay Merchant validation depending on configuration

marcotranchino:
With this change, we are introducing a new environment variable to check
whether the pay-frontend application should use `axios` for the Apple Pay
Merchant Validation.

By default the new variable will not exist on any environment, therefore
pay-frontend will keep using `request retry` to operate the Apple Pay Merchant
Validation.

When the new variable is created and set to true, then pay-frontend will use
`axios`, with or without the use of HttpsProxyAgent, in order to operate the
Apple Pay Merchant Validation.

This will allow us to safely deploy this change to all environments including
production, and crucially it will allow us to test the Apple Pay Merchant
Validation using `axios` on the Test environment only, by creating the new
environment variable there and setting it to 'true'.

Further work will be needed to remove the old code and the use of request
retry once the validation will be successfully operated with `axios` in all
environments.

Note that for simplicity the old tests for the Apple Pay Merchant Validation
using `request retry` have been saved in a new file called
`old-merchant-validation.controller.test.js`, so that it will be easy to simply
delete this file once the current way using request retry will be fully
discontinued.

Further information in Jira.

## screenshots of logs when request retry is used (default)

with request retry

## screenshots of logs when axios is used (to be enabled on test environment)

with axios

### DIFF
--- a/app/controllers/web-payments/apple-pay/merchant-validation.controller.js
+++ b/app/controllers/web-payments/apple-pay/merchant-validation.controller.js
@@ -66,6 +66,8 @@ module.exports = async (req, res) => {
     {
       url: url,
       method: 'post',
+      cert: merchantIdentityVars.cert,
+      key: merchantIdentityVars.key,
       headers: { 'Content-Type': 'application/json' },
       data: {
         merchantIdentifier: merchantIdentityVars.merchantIdentifier,


test/controllers/web-payments/apple-pay/merchant-validation.controller.test.js
```diff
@@ -12,172 +14,226 @@ const stripeCertificate = 'A-STRIPE-CERTIFICATE'
...
+          merchantIdentifier: worldpayMerchantId,
+          displayName: 'GOV.UK Pay',
+          initiative: 'web',
+          initiativeContext: merchantDomain
+        },
+        httpsAgent: sinon.match.instanceOf(HttpsProxyAgent)
+      }))
+
+      const httpsAgentArg = axiosStub.getCall(0).args[0].httpsAgent
+      // sinon.assert.match(httpsAgentArg.options, { // not working locally for me
```
hjvoid, line 112: Is this still needed?

hjvoid (approved):
Looks great - nice one 💪

PR #3881, opened 2024-07-31
PP-12546 generate tokens from session secret over charge secret

nlsteers:
## WHAT

this change is part of a wider series that are being merged in sequence to prevent the need for a downtime deployment

- `setCsrfSecret` middle-ware has been deployed and any currently in-flight payments at the time of this deployment should have a top level session csrf secret
- modify the `csrfTokenGeneration` middle-ware to use the top level session csrf secret instead of the charge level csrf secret
- remove redundant log line from `csrfCheck` middle-ware, `fetchAndValidateChargeId` is called ahead of the removed log line and already provides all the necessary logging detail

### DIFF
--- a/app/middleware/csrf.js
+++ b/app/middleware/csrf.js
@@ -20,16 +20,14 @@ exports.csrfSetSecret = (req, _, next) => {
 }
 
 exports.csrfTokenGeneration = (req, res, next) => {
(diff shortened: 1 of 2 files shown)


DomBelcher (approved):
LGTM

PR #3885, opened 2024-08-02
PP-12853 Validate Apple Pay passing HttpsProxyAgent to Axios post

marcotranchino:
With this change, we are making a new attempt at using Axios with HttpsProxyAgent to operate an Apple Pay Merchant validation when there is a proxy, which is the configuration in place for all of our environments on AWS.

The Apple Pay validation with Axios is switched on only on the Test environment, not on production, therefore this change is safe to deploy because it does not change the validation with `request retry` which is used for production.

This new version passes the HttpsProxyAgent within the post method, in the third parameter, together with headers.

Further information in the Axios documentation[1], which unfortunately does not have examples with HttpsProxyAgent, and in the Jira ticket[2].

[1]
https://axios-http.com/docs/post_example

[2]


SandorArpa (approved):
LOKTM

PR #3891, opened 2024-08-09
Use `hpagent` with `axios` for Apple Pay Merchant Validation with egress proxy

marcotranchino:
With this change, we have update the Apple Pay Merchant Validation implementation in order to use `hpagent` and `axios` in the presence of an egress proxy, as it's the case on our AWS environments.

This is needed because we want to remove the use of `requestretry`, however `axios` has a problem preventing it from working with an egress proxy[1].

For this reason, we need to use an HttpsProxyAgent with it.

We would want to use `https-proxy-agent`, however it has its own problem[2].

While we wait for these issues to be fixed, we can use `hpagent` which has been tested and works well with an egress proxy.

Further information in the JIRA ticket[3].

[1]
https://github.com/axios/axios/issues/4531

[2]
https://github.com/TooTallNate/proxy-agents/pull/235

[3]

Co-authored-by: Jonathan Harden <jonathan.harden@digital.cabinet-office.gov.uk>
Co-authored-by: Dominic Belcher <dominic.belcher@digital.cabinet-office.gov.uk>
Co-authored-by: Marco Tranchino <marco.tranchino@digital.cabinet-office.gov.uk>

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -76,6 +76,10 @@
       "path": "detect_secrets.filters.allowlist.is_line_allowlisted"
     },
     {
+      "path": "detect_secrets.filters.common.is_baseline_file",
+      "filename": ".secrets.baseline"
+    },
+    {
       "path": "detect_secrets.filters.common.is_ignored_due_to_verification_policies",
       "min_level": 2
     },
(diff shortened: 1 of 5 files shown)


DomBelcher (approved):
Awesome

PR #3892, opened 2024-08-09
Remove code path for Apple Pay validation with request retry

marcotranchino:
With a recent change[1], we have enabled Apple Pay Merchant Validaton with
`axios` and `hpagent`.

The change has been successfully verified on the Test environment, where the
code path with `axios` was switched on using an environment variable[2].

Now that the code path with `axios` has been switched ON also on Production,
then we are now able to stop using `requestretry` in pay-frontend by
completely deleting the code path using it for the Apple Pay Merchant
Validation.

Note that whilst working on this code, we have also improved the Pay CLI, and
that now it is possible to run Pay locally --with-egress-proxy. For this
reason, we are also removing the code path using `axios` without a proxy URL,
since it is not production like and it is not needed for local development
either.

Further information in JIRA[3].

[1]
https://github.com/alphagov/pay-frontend/pull/3891

[2]
https://github.com/alphagov/pay-infra/pull/4913

[3]


hjvoid (dismissed):
LGTM 🍾

app/controllers/web-payments/apple-pay/merchant-validation.controller.js
```diff
@@ -66,92 +63,59 @@ module.exports = async (req, res) => {
...
-      const data = {
-        merchantIdentifier: merchantIdentityVars.merchantIdentifier,
-        displayName: 'GOV.UK Pay',
-        initiative: 'web',
-        initiativeContext: process.env.APPLE_PAY_MERCHANT_DOMAIN
-      }
-
-      try {
-        const response = await axiosInstance.post(url, data, { headers: { 'Content-Type': 'application/json; charset=utf-8' } })
+  const options = {
```
DomBelcher, line 66: This can be scoped to the `else` block where it's used in the request

app/controllers/web-payments/apple-pay/merchant-validation.controller.js
```diff
@@ -66,92 +63,59 @@ module.exports = async (req, res) => {
...
-      }
+    const data = {
+      merchantIdentifier: merchantIdentityVars.merchantIdentifier,
+      displayName: 'GOV.UK Pay',
+      initiative: 'web',
+      initiativeContext: process.env.APPLE_PAY_MERCHANT_DOMAIN
     }
 
+    try {
+      const response = await axiosInstance.post(url, data, { headers: { 'Content-Type': 'application/json; charset=utf-8' } })
```
DomBelcher, line 92: In theory we shouldn't need to use a separate axios instance, doing `axios.post(url, data, config)` should work, so long as the `config` object has the `httpsAgent` set as the proxy agent and `proxy: false` set.

DomBelcher: Maybe something to leave for a later PR though so it can be tested separately

app/controllers/web-payments/apple-pay/merchant-validation.controller.js
```diff
@@ -66,92 +63,59 @@ module.exports = async (req, res) => {
...
-        })
-        logger.info('Apple Pay session via request retry failed', body)
-        return res.status(500).send(body)
-      }
-      logger.info('Apple Pay session successfully generated via request retry')
-      res.status(200).send(body)
-    })
+    logger.info('Generating Apple Pay session via axios and https proxy agent (hpagent) (NO PROXY)')
+    try {
+      const response = await axios(options)
```
DomBelcher, line 108: Is this path ever going to be hit? It shouldn't work as there's no `proxy: false` set in the options - axios will still try to use it's internal proxy and fail

DomBelcher (changes requested):

DomBelcher (approved):
LGTM

PR #3894, opened 2024-08-14
PP-12853 Remove `request` from pay-frontend

marcotranchino:
After having implemented Apple Pay Merchant Validation with `axios`, we can now fully remove `request` from pay-frontend.

Further information in Jira.

https://payments-platform.atlassian.net/browse/PP-12853

PR #3896, opened 2024-08-14
Remove the `request` related Cypress override

marcotranchino:
After having removed `request` from pay-frontend[1], we can now also remove the Cypress override needed for `request`.

Further info in Jira[2].

[1]
https://github.com/alphagov/pay-frontend/pull/3894

[2]


SandorArpa (approved):
LOKTM

PR #3901, opened 2024-09-05
[WIP] PP-13119 fix correlation id

SandorArpa:
## WHAT

The Node apps were updated to use middleware that used the async_hooks - to help with logging. However, there is an issue where the correlation is not being set correctly and not being passed to the backend call.

- fix correlation id issue


iqbalgds:
Closing this as it will be replaced with this:
https://github.com/alphagov/pay-frontend/pull/3916

PR #3908, opened 2024-09-23
Moved Chokidar dependency to development as it's not used in Prod

nataliecarey:
This resolves the braces triage.

## WHAT
There's a potential vulnerability in `braces` which is used by Chokidar and Style Lint.  The vulnerability is related to user input so we aren't affected as we don't put user input through either of these tools.

To be sure that we're not in any danger I've moved Chokidar to a dev dependency - I presume it should have always been a dev dependency.  This means it won't go live and there's zero chance that user input will be put through it.

To confirm this you can run the command `npm ls braces` and `npm ls --omit=dev braces` against the codebase with and without this change or check the `node_modules` directory after running `npm install --omit=dev` on a clean clone.


nlsteers (approved):
👍🏻

PR #3916, opened 2024-10-16
PP-13119 Fix logging of correlation ID to log messages

iqbalgds:
- Remove references to `async local storage`.
- No longer using this as that will be a big change.
- Instead continue to save logging context fields to the request instead.
- Remove all references to `async local storage` - including the `request-context` file.
- Fix bug where the correlation id was not getting logged and not sent as a header to backend services.
  - When configuring a client that calls the backend - we now pass the correlation id and save it for each request.
  - Pass the correlation id to the backend as a header.
  - Log the correlation id as part of logging messages.
- Remove logging messages that are no longer required as the `pay-js-commons/axios-base-client` will do the logging instead.


hjvoid (approved):
Looks cool to me - thanks for the walkthrough! 👍

PR #3920, opened 2024-10-18
PP-12828 Update logging in connector client to only log 5xx responses as errors

DomBelcher:
## WHAT
 - Update logging on `GET` requests to Connector
   - this will only log `5xx` responses as errors
   - all other non-`200` responses will be logged as warnings


nlsteers (approved):
LGTM

PR #3924, opened 2024-10-24
PP-11883 Update cancel link to be a button

iqbalgds:
- The cancel link on the payment and confirm pages appears like a link but is semantically a button.
- This can cause confusion for users using assitive tech.
- This PR changes the link so that it is a secondary action button.
- Also got rid of SASS that is no longer required.

New styling below:

image


SandorArpa (dismissed):
LGTM

SandorArpa (approved):
LGTM

PR #3937, opened 2024-12-27
PP-13239 Add Cypress tests for custom branding

iqbalgds:
- Add Cypress tests for custom branding
  - Test white background with dark text.
  - Test purple background with white text.
- pay-frontend caches responses from admin users >   find service.
  - So each test needs to use a different gateway id or it returns the cached custom branding data.


SandorArpa (approved):
LGTM

PR #3949, opened 2025-02-03
PP-12475 - Remove HTML5 polyfill and IE8 stylesheet references

iqbalgds:
- The HTML5 polyfill and IE8 stylesheet files have been removed previously.
- This commit removes the references to those files as they are no longer required
  - This is because we no longer provide full support to older that require these files.


SandorArpa (approved):
LGTM

PR #3951, opened 2025-02-11
PP-11093: Remove unused dev dependencies from pay-js-metrics

oswaldquek:
[Dev dependencies from pay-js-metrics](https://github.com/alphagov/pay-js-metrics/blob/main/package.json#L28) shouldn't be included.


DomBelcher (approved):
LGTM

PR #3963, opened 2025-04-01
Update to use latest js commons

hjvoid:
## WHAT
Updating to use latest pay-js-commons after custom branding update.


SandorArpa (approved):
LOKTM

PR #3983, opened 2025-06-23
PP-14256 Add rebrand error page test to rebrand test suite

iqbalgds:
- A test to check the error pages containing the new branding was added to the code base.
- This PR is just add this test to the rebrand test suite so that it is only run when testing the new branding.


SandorArpa (approved):
LOKTM

PR #3986, opened 2025-06-27
PP-14152- Remove rebranding flag and related code.

iqbalgds:
- New branding is now on PROD.
- Remove feature flag code - have the new branding changes hardcoded in.
- Update existing Cypress to only check for new branding.
- Remove the specific rebranding Cypress tests as the existing Cypress tests will be updated instead.
- Remove the Github action step to run the specific Cypress rebranding tests
  - This is no longer required as the existing tests will be updated.


SandorArpa (approved):
LGTM

PR #3987, opened 2025-07-07
PP-14027 Remove forward proxy code

kbottla:
## WHAT
- Removed usages of FORWARD_PROXY_URL from healthcheck and related code, as clients no longer use forward proxy.


danworth (approved):
Looks good 🚢

PR #4003, opened 2025-08-06
PP-8287 Remove publicly visible webpack stats page from Frontend

AbdirizakIdris:
## WHAT

 - ensure `stats.html` is no longer copied into `govuk_modules` during the build


SandorArpa (approved):
LGTM

PR #4008, opened 2025-08-29
PP-14362- Added a listener to address-country to make sure it runs validation on change

Rafiq-Gasmi:
## ISSUE
When the Country is changed it does not run validation on the form fields, meaning errors are not cleared or added. This is due to the address-country listener not created due to address-country not loaded on DOM when Init() runs

## WHAT
Updated a broken link from the read me.
Added a DOM watcher to see when address-country exists so checkValidation can run.
Added a test to check that the MutationObserver attaches a listener to address-country when it is added to the DOM, and then disconnects itself.

## HOW 
_Steps to test or reproduce:_
1. Go to payment page
3. Scroll to address fields
5. Put post code in first - anything in a non UK format. Eg. 00000 
7. See validation correctly reject this (because country is set to united kingdom) 
9. Change United Kingdom in ‘Country or territory’ to Hong Kong
11. See that the validation is still causing an error  (this is the bug) 
13. Tab through 'Country or territory’ and post code to remove the validation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
     "helmet": "^7.1.0",
     "hpagent": "^1.2.0",
     "i18n": "0.15.x",
+    "jsdom": "^26.1.0",
     "lodash": "4.17.x",
     "mailcheck": "^1.1.1",
     "memory-cache": "^0.2.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -87,6 +87,7 @@
     "helmet": "^7.1.0",
     "hpagent": "^1.2.0",
     "i18n": "0.15.x",
+    "jsdom": "^26.1.0",
```
nlsteers, line 90: is this just a test dep? If so should be in dev deps

nlsteers (approved):
nice work, ship it

PR #4060, opened 2026-05-08
PP-15362 Switch to plain alpine tag

benjamb:
## WHAT
Switch to plain alpine tag so minor alpine upgrades are performed automatically.

## HOW 
_Steps to test or reproduce:_


SandorArpa (commented):
LOKTM

SandorArpa (approved):

PR #4072, opened 2026-06-09
PP-15406 upgrade to pact 16

ssrome:
## WHAT
- Upgraded NPM modules
- Updated tests to work with pact 16
- Updated how pact test are published to the pact 16 method


iqbalgds (approved):

PR #4080, opened 2026-06-26
PP-15417 Collect additional browser info for Adyen payments

kbottla:
## WHAT
- Collects additional browser info for Adyen non-MOTO payments
- Repurposed existing `epdq-3ds2.js` for Adyen, as we no longer support ePDQ, and the script required is similar for Adyen
- Adds new `jsEnabled` hidden field with default value as `false`. This is so that a value is sent to the Connector when JavaScript is disabled. We need to send the value  `javaScriptEanbled: false` explicitly to Adyen (Default is `true` when nothing is sent)
     - The field will be set to `true` if JavaScript is enabled in the browser.


AndrzejProchyra (approved):
LGTM!

PR #4095, opened 2026-08-06
PP-15669 - Made Apple Pay merchant-validation work with Adyen payment provider

GabrielBoardman-MadeTech:
- Added Adyen case to getApplePayMerchantIdentityVariables
- Replaced if block with switch statement


alexbishop1 (approved):

PR #4098, opened 2026-08-13
PP-15730 Update to Cypress 15

DomBelcher:
## WHAT
 - Upgrade Cypress version to `^15.20.0`
- Fix failing tests caused by change to cy.stub call signature (see https://docs.cypress.io/app/references/migration-guide#cystub-3-argument-signature-removed for more info)
- This should address test failures in #4091


christiandutton-mt (approved):
LGTM!

PR #4100, opened 2026-08-19
PP-15720 Google pay normalisation adds additional adyen browser info

hjvoid:
- Added Adyen browser detail to `paymentInfo` object if the `paymentProvider` is Adyen
- Added `performDeviceDataCollectionForAdyen` method in `normalise-google-pay-payload` to append the values mentioned above
- Added to the `normalise-google-pay-payload` test cases

### DIFF
--- a/app/assets/javascripts/browsered/collect-additional-browser-info-adyen.js
+++ b/app/assets/javascripts/browsered/collect-additional-browser-info-adyen.js
@@ -1,5 +1,8 @@
 'use strict'
 
+const { keysToCamelCase } = require('../../../utils/key-camelizer')
+const { getBrowserInfo } = require('./web-payments/helpers')
+
 const init = () => {
   if (window.Charge.collect_additional_browser_info_adyen) {
     addAdditionalInformation()
@@ -8,26 +11,11 @@ const init = () => {
 
 const addAdditionalInformation = () => {
   document.getElementById('jsEnabled').value = 'true'
-  if (typeof navigator.language === 'string') {
-    appendHiddenInputToForm('jsNavigatorLanguage', navigator.language)
+  const browserInfo = keysToCamelCase(getBrowserInfo())
+
+  for (const [key, value] of Object.entries(browserInfo)) {
+    appendHiddenInputToForm(key, value)
   }
-
-  if (window.screen) {
-    if (typeof window.screen.colorDepth === 'number') {
-      appendHiddenInputToForm('jsScreenColorDepth', window.screen.colorDepth)
-    }
-
-    if (typeof window.screen.height === 'number') {
-      appendHiddenInputToForm('jsScreenHeight', window.screen.height)
-    }
-
-    if (typeof window.screen.width === 'number') {
-      appendHiddenInputToForm('jsScreenWidth', window.screen.width)
-    }
-  }
-
-  const date = new Date()
-  appendHiddenInputToForm('jsTimezoneOffsetMins', date.getTimezoneOffset())
 }
 
 const appendHiddenInputToForm = (name, value) => {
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/browsered/web-payments/google-pay.js
```diff
@@ -112,6 +112,12 @@ const processPayment = paymentData => {
   toggleSubmitButtons()
   showSpinnerAndHideMainContent()
 
+  if (payment_provider === 'adyen') { // eslint-disable-line camelcase
```
kbottla: Any reason if conditions are nested?

```suggestion
  if (payment_provider === 'adyen' && Charge.collect_additional_browser_info_adyen === true ) { // eslint-disable-line camelcase
```

app/assets/javascripts/browsered/web-payments/google-pay.js
```diff
@@ -163,6 +169,25 @@ const createGooglePaymentRequest = () => {
   return new PaymentRequest(methodData, details, options)
 }
 
+const performDeviceDataCollectionForAdyen = (paymentData) => {
+  paymentData.js_enabled = true
+  if (navigator.language === 'en-GB') {
```
kbottla: We don't need to check for specific values and set them. If so, for a browser with different property values, the fields will always be undefined.

We can pass values as-is to the frontend. So, this will become

```
paymentData.js_enabled = true
paymentData.js_navigator_language = navigator.language
paymentData.js_screen_color_depth = window.screen.colorDepth
paymentData.js_screen_height = window.screen.height
paymentData.js_screen_width = window.screen.width
paymentData.js_timezone_offset_mins = new Date().getTimezoneOffset()
```

app/assets/javascripts/browsered/web-payments/google-pay.js
```diff
@@ -163,6 +169,25 @@ const createGooglePaymentRequest = () => {
   return new PaymentRequest(methodData, details, options)
 }
 
+const performDeviceDataCollectionForAdyen = (paymentData) => {
```
kbottla: We are collecting browser information in multiple places, including here and in collect-additional-browser-info-adyen.js.

It would be good to add a common method to collect-additional-browser-info-adyen.js that returns browser data for use by both the card details page and Google Pay.

kbottla: paymentData is the data we get from Google Pay, and for Worldpay device data collection, we add separate fields to the requestBody instead of appending paymentData. 

Should we add a separate JSON object for browser info so it stays separate from the actual Google Pay payment data?

app/assets/javascripts/browsered/web-payments/google-pay.js
GabrielBoardman-MadeTech: That makes sense, thank you!

app/assets/javascripts/browsered/web-payments/google-pay.js
GabrielBoardman-MadeTech: which file is the card details page? Is it `form-cards-type.js`?

kbottla: For the details page, JS method called is https://github.com/alphagov/pay-frontend/blob/26243da0120be74d222b75d2f0404ee964ce0a68/app/assets/javascripts/browsered/collect-additional-browser-info-adyen.js#L5

GabrielBoardman-MadeTech: We can add an additional json to `paymentData`, called `browser_info` instead of appending the data directly to the paymentData.

app/assets/javascripts/browsered/collect-additional-browser-info-adyen.js
```diff
@@ -8,26 +11,11 @@ const init = () => {
 
 const addAdditionalInformation = () => {
   document.getElementById('jsEnabled').value = 'true'
-  if (typeof navigator.language === 'string') {
```
joshuapook-MT, line 11: Do you not need these type checks anymore?

GabrielBoardman-MadeTech (dismissed):

kbottla (approved):
LGTM!

PR #4102, opened 2026-08-20
PP-15719 send request to connector on adyen specific api route

GabrielBoardman-MadeTech:
- Add functionallity to allow payment-auth-request.controller to send data to the correct connector route
- Add Adyen Google Pay methods to connector.client

### DIFF
--- a/app/controllers/web-payments/google-pay/normalise-google-pay-payload.js
+++ b/app/controllers/web-payments/google-pay/normalise-google-pay-payload.js
@@ -94,8 +94,8 @@ module.exports = (req, paymentProvider) => {
     ip_address: userIpAddress(req)
   }
 
-  const paymentData = keysToSnakeCase(JSON.parse(payload.paymentResponse.details.paymentMethodData.tokenizationData.token))
-
+  const token = payload.paymentResponse.details.paymentMethodData.tokenizationData.token
+  const paymentData = keysToSnakeCase(JSON.parse(token))
   delete payload.paymentResponse.details.paymentMethodData
 
   switch (paymentProvider) {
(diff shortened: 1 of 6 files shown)


app/controllers/web-payments/payment-auth-request.controller.js
```diff
@@ -61,3 +65,26 @@ module.exports = (req, res, next) => {
...
+    .catch(err => {
+      logger.error('Error while trying to authorise Google Pay Adyen payment', {
+        ...getLoggingFields(req),
+        error: err
+      })
+      res.status(200)
+      // Always return 200 - the redirect handles the error
+      res.send({ url: `/handle-payment-response/google/${chargeId}` })
+    })
+}
```
kbottla: Is it possible to extract this block into a separate method, as it is the same for both Adyen and other payment providers

GabrielBoardman-MadeTech: Hi Krishna, we removed this in an [earlier commit](https://github.com/alphagov/pay-frontend/compare/a31fdc019f5c14f82fde0ef2a30de10d7c745e6f..16ec0b10fe654b24e637703736d1095c8502847e), unsure why thats reverted when i rebased :/
Will fix this now

app/controllers/web-payments/google-pay/normalise-google-pay-payload.js
```diff
@@ -94,8 +94,8 @@ module.exports = (req, paymentProvider) => {
     ip_address: userIpAddress(req)
   }
 
+  const token = payload.paymentResponse.details.paymentMethodData.tokenizationData.token
   const paymentData = keysToSnakeCase(JSON.parse(payload.paymentResponse.details.paymentMethodData.tokenizationData.token))
```
kbottla:
```suggestion
  const paymentData = keysToSnakeCase(JSON.parse(token))
```

kbottla (dismissed):
LGTM!

Added a minor comment but not blocking

kbottla (approved):

PR #4106, opened 2026-09-01
PP-15812 Fix Adyen payment normalisation error

alexbishop1:
Fix error normalising entered payment data when making an Adyen payment introduced by commit 81c1c66f656017a051303a7115d47b5dbe84e9fb.

The error was caused by the client-side JavaScript adding a duplicate hidden form field with the name `jsEnabled`, which Node.js handles by combining the two values into an object, which means it’s not a string when later code tries to trim it.

Fix the problem by making sure that there is only one `jsEnabled` hidden form field (or `js_enabled` JSON property in the case of Google Pay payments). Enhance the existing Cypress test that tests these hidden form fields by making it submit the form to ensure they’re actually processed correctly on the server side.


hjvoid (approved):

PR #4110, opened 2026-09-08
PP-14499 Fix inconsistency in readme

alexbishop1:


hjvoid (approved):
I’m confident that this change does not require further review and meets the secure coding requirements of [GOV.UK](http://gov.uk/)  Pay

PR #4116, opened 2026-09-16
PP-15923 - Update getBrowserInfo() to use camel case for consistency

GabrielBoardman-MadeTech:
- Update `getBrowserInfo()` to use camel case for consistency
- Update relevant usages and test


alexbishop1 (approved):

PR #4117, opened 2026-09-17
PP-15927 - Changed browserInfo to be on the same level as paymentResponse in requestBody

GabrielBoardman-MadeTech:
What we did:
- Unnest browserInfo from paymentResponse
- Correct where we pull the browserInfo fields from in normalise-google-pay-payload
- Wrap the extraction of browserInfo fields in `if` blocks
- Modify appropriate tests


hjvoid (approved):
I’m confident that this change does not require further review and meets the secure coding requirements of [GOV.UK](http://gov.uk/)  Pay

